PR #13, opened 2023-06-06
feat: rename osumemory-ts to tosu

cyperdark:
and repository too


KotRikD:
Okay, well actually I calling it tosu or tsosu, so yeah good thing

PR #14, opened 2023-06-06
fix: optimize beatmap parsing

kionell:
Currently, you are parsing unused beatmap sections, including storyboards. 
Sometimes, storyboards can be embedded directly into .osu files, causing you to spend extra time parsing beatmap.


KotRikD:
LGTM

PR #19, opened 2023-12-08
WIP: modules language

KotRikD:
Idea to make memory entities with lanauge syntax, like it was in gosu
`[Something + 0x4]`


KotRikD:
a

KotRikD:
Need to re-think on that, probably not good idea

PR #38, opened 2024-02-13
Switch from `find-process` to rust alternative

cyperdark:
Pros:
- -1 dependency
- from `400ms` to `20ms` (on my machine)

<br>

_p.s. Need linux testing_


packages/find-process/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "find-process"
+version = "0.0.1"
+authors = ["ck"]
+license = "ISC"
+edition = "2018"
```
KotRikD, line 6: 2018?

packages/find-process/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "find-process"
+version = "0.0.1"
+authors = ["ck"]
+license = "ISC"
```
KotRikD, line 5: maybe MIT?

KotRikD (changes requested):

KotRikD (approved):

PR #72, opened 2024-03-01
feat: Improve GitHub Workflow (deploy.yml)

GabuTheDev:
This PR changes the `deploy & build` workflow behaviour by restricting action path to client-related files. (+ some reformatting)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,12 +1,14 @@
-name: Build and Deploy
+name: build & deploy
 
 on:
   push:
-    branches:
-      - 'master'
+    branches: [ "master" ]
     tags:
-      - 'v*'
-  pull_request:
+      - "v*"
+    paths:
+      - "packages/**"
+      - "package.json"
+      - "pnpm-lock.yaml"
 
 concurrency:
   group: build-${{ github.sha }}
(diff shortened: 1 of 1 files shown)


GabuTheDev:
May need testing before merging to `master`.
I'm not familiarized with GitHub Actions.

.github/workflows/deploy.yml
```diff
@@ -19,59 +21,55 @@ jobs:
...
-          pnpm run build
+      - name: 🚀 - Upload artifacts for windows
+        if: matrix.os == 'windows-latest'
+        uses: actions/upload-artifact@v3
+        with: {
+          name: "tosu-${{ matrix.os }}-${{ steps.set-pr-sha.outputs.SHORT_PR_SHA || github.ref_name }}",
+          path: packages/tosu/dist/tosu.exe
+        }
 
-      # - name: Upload Artifact (Linux)
```
KotRikD, line 58: Please return commentaries, currently Linux & Mac aren't supported, due to library

KotRikD (changes requested):

KotRikD:
And probably need add to paths `.github/workflows/**/*` to test any changes in CI as in your PR

GabuTheDev:
> And probably need add to paths `.github/workflows/**/*` to test any changes in CI as in your PR

I do not think it's needed. There's no need for a build or deploy when editing `.github/` since the files inside this directory are steps for the GitHub CI that are being run on their server, not in the client (.exe).

If you want to add testing to PRs you should add a separate workflow that specifically does that testing.

.github/workflows/deploy.yml
GabuTheDev: > Please return commentaries, currently Linux & Mac aren't supported, due to library
They wouldn't have ran anyways since the condition is not met.

GabuTheDev: > Please return commentaries, currently Linux & Mac aren't supported, due to library
They wouldn't have ran anyways since the condition is not met.

GabuTheDev: > Please return commentaries, currently Linux & Mac aren't supported, due to library

They wouldn't have ran anyways since the condition is not met.

GabuTheDev:
I can do a separate workflow specifically for testing PR changes (in a separate PR) if needed.

PR #109, opened 2024-03-19
fix: fix tournament manager process order

KotRikD:


KotRikD:
Thanks for idea of `getInstance` to @ILW8

PR #146, opened 2024-05-27
Fix bug with retrieving beatmap data due to improper URL encoding

thenerdie:
I'm using tosu in a personal project of mine, and when I tried to access map data for certain songs, it was giving me a 404 error. Funny. As it turns out, such directory names contained URL reserved characters such as "#", and since the server was just running `decodeURI()` on the whole URL, it wasn't treating such reserved characters LITERALLY.

Take this directory name for example: `1032511 Hino Isuka - #be_fortunate`

Obviously, the `#` is part of the song name, thus we want to treat it as a literal part of the resource name, instead of as a page locator tag. So I implemented `encodeURIComponent()` on each part of the URL, so that it would treat the # literally, encoding it as %23. :3

### DIFF
--- a/packages/server/utils/directories.ts
+++ b/packages/server/utils/directories.ts
@@ -23,7 +23,7 @@ export function directoryWalker({
 }) {
     let cleanedUrl;
     try {
-        cleanedUrl = decodeURI(pathname);
+        cleanedUrl = decodeURIComponent(pathname);
     } catch (error) {
         res.writeHead(404, {
             'Content-Type': getContentType('file.txt')
@@ -113,9 +113,10 @@ export function readDirectory(
 
         const html = folders.map((r) => {
             const slashAtTheEnd = getContentType(r) === '' ? '/' : '';
+
             return `<li><a href="${
                 url === '/' ? '' : url
-            }${r}${slashAtTheEnd}">${r}</a></li>`;
+            }${encodeURIComponent(r)}${slashAtTheEnd}">${r}</a></li>`;
         });
 
         return callback(


packages/server/utils/directories.ts
```diff
@@ -23,7 +23,10 @@ export function directoryWalker({
 }) {
     let cleanedUrl;
     try {
-        cleanedUrl = decodeURI(pathname);
+        cleanedUrl = pathname
```
cyperdark, line 26: you can just simly use `decodeURIComponent`, no need of doing this amount of garbage, and also test your code with different approaches first

image
image

packages/tosu/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "tosu",
-    "version": "2.7.0",
+    "version": "3.0.0",
```
cyperdark, line 3: DO NOT CHANGE VERSIONS

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "author": "Mikhail Babynichev",
   "license": "GPL-3.0",
-  "version": "2.7.0",
+  "version": "3.0.0",
```
cyperdark, line 4: DO NOT CHANGE VERSIONS

CHANGELOG.md
```diff
@@ -2,6 +2,276 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## 3.0.0 (2024-05-27)
```
cyperdark, line 5: remove it

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
cyperdark, line 1: revert this file too

cyperdark (changes requested):

packages/server/utils/directories.ts
thenerdie: `decodeURIComponent` encodes the slash character. I can't shove the entire pathname in there. The code I splits the full URL by slashes "blahblahblah#/blahblah" and encodes each chunk with `decodeURIComponent`, and this works just fine. Not sure what the issue is here.

KotRikD: Hmmm, `decodeURI` seems shouldn't be used here, because `decodeURI/encodeURI` is designed to be used in full path URI's. Let's talk about solution with `decodeURIComponent`, @thenerdie can you provide cases, where standard `decodeURIComponent` fails and you needed to use map function?

thenerdie: Fixed. I was braindead coding last night LOL

thenerdie:
All requested changes should have been handled.

KotRikD:
LGTM

PR #150, opened 2024-06-08
fix bad map id momentarily being sent when hitting panic button in tclient 

ILW8:
follow-up to #109

tosu has regressed compared to gosu wherein it still occasionally returns the wrong map ID when certain actions occur on the tournament client (e.g. pressing panic button when gameplay starts and client #0 hasn't loaded in).
A number of tournament overlays update pick/bans using current map ID, and this flip-flopping of map IDs can mess up autopick. Of course, it's fairly trivial to fix this on the overlay side, but a regression is a regression.

### DIFF
--- a/packages/tosu/src/api/utils/buildResult.ts
+++ b/packages/tosu/src/api/utils/buildResult.ts
@@ -1,3 +1,4 @@
+import { wLogger } from '@tosu/common/utils/logger';
 import path from 'path';
 
 import {
@@ -81,6 +82,8 @@ export const buildResult = (instanceManager: InstanceManager): ApiAnswer => {
         50: resultsScreenData.Hit50,
         0: resultsScreenData.HitMiss
     };
+
+    wLogger.info(`[${osuInstance.pid}] current mapid ${menuData.MapID}`);
 
     return {
         settings: {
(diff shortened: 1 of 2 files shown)


cyperdark:
can you showcase the issue? video will be cool @ILW8

ILW8:
Sure, I'll try to replicate it when another streamer can take a match (probably sunday). It happens infrequently enough that it's hard to reproduce, but frequently enough that it keeps you on the lookout while streaming.

ILW8:
Well it just happened, what a coincidence! Unfortunately I don't have a recording of the logs, but you can see on stream that the "picked by" switched from China to United States after hitting the panic button in tclient.
https://www.twitch.tv/videos/2166479543?t=00h10m57s

To illustrate the issue better, assume that the currently picked map is map id `112233`. When entering gameplay and then pressing Panic, here's what the ws messages might look like:
- map id `112233`
- map id `112233`
- map id `112233`
- map id `112233`
- map id `445566`
- map id `445566`
- map id `112233`
- map id `112233`
where `445566` is any arbitrary other map ID.
When the messages go back and forth between `112233` and `445566`, the overlay thinks `112233` got picked again and marks it as picked by the other team.

You can find the overlay code at https://github.com/shdewz/4wc-stream-overlay/blob/main/mappool/index.js. The same map auto-pick flow worked fine for 3WC which used gosumemory

PR #162, opened 2024-07-08
feat: add rust memory reader

TheEZIC:
- Migrate memory reader to rust lang;
- Create an abstraction under memory reading in tosu package;
- Replace PROCESS_POWER_THROTTLING_IGNORE_TIMER_RESOLUTION constant with it's value because my compiler can't find this variable somehow and fails building;
- Install @types/adm-zip to @tosu/common package as dev dependency.

Not supported methods yet:
- disablePowerThrottling;
- getProcesses;
- scanSync;
- scan;
- scanAsync


KotRikD:
1. conflicts
2. @cyperdark can you perform performance test?

KotRikD (changes requested):
1. conflicts
2. @cyperdark can you perform performance test?

TheEZIC:
fixed

TheEZIC:
If you want I can try to add memory reader engine select into setting

TheEZIC:
I found one bug with big addresses which above i32::MAX when passing them from JS to Rust. Rust casts them as i32::max

Here is how CPP casts
image

Here is how Rust casts
image

cyperdark:
At this current state it's slower, need more precise tests to determine if it's faster/same or slower

timings parser: https://github.com/cyperdark/tosu-timings-parser
p.s. Making it draft for now


1721031314_0ncAj4oUvc
1721031311_txWUwVIFfW
1721031308_3lP8DnfqBF
1721031305_IXhkQywXw9
1721031289_bHZtDQ4dEL
1721031283_0KSx9nj4Vb
1721031280_BOh3KKHQZE
1721031275_zI3cbyHZ02
1721031263_Hjya5VIP40

TheEZIC:
Here is my benchmarks on same maps:

[rust.txt](https://github.com/user-attachments/files/16233342/rust.txt)
[cpp.txt](https://github.com/user-attachments/files/16233344/cpp.txt)

TheEZIC:
RIP

PR #163, opened 2024-07-08
feat: add KeyOverlay to tournament clients

### LINKED ISSUE #160: Support KeyOverlay on tournament clients
Is it possible to add KeyOverlay support for each tournament client?

TheEZIC:
Implements KeyOverlay to tournament clients. I didn't check it in real case, because I don't have osu! supporter.

Resolves #160


cyperdark:
KeyOverlay shouldn't be there, and also we have some internal disscussion about it, mainly because kotrik want to have it differently

cyperdark:
also no types

TheEZIC:
> also no types

```ts
{ instanceManager }: { instanceManager: any }
```

PR #168, opened 2024-07-17
Update delay & HitErrors & improves

cyperdark:
- credits
- Switching to old method of pushing HitErrors
- Remove delay for non-tournament clients
- Prevent updating if one of values `not-ready`
- adjust spam factor for HitErrors
- Use `UTC+0` for `createdAt` on result screen
- fix `hitError` issue, and `null pointer`

### DIFF
--- a/packages/server/assets/homepage.html
+++ b/packages/server/assets/homepage.html
@@ -56,7 +56,7 @@
     </div>
   </main>
   <footer>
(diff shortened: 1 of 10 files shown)


cyperdark:
also fixed memory leak for `calculate/pp` and optimized calculation for current map
image

KotRikD (changes requested):

PR #176, opened 2024-08-08
fix: read addresses as uint32 instead of int64

xxCherry:
this pr fixes issues when some addresses were negative when converted from int64 to unsigned version
so this just reads all addresses coming from node as uint32 instead, this shouldn't be an issue unless we plan to add lazer support


KotRikD (approved):

cyperdark:
ILOVEMANIA <#####


https://github.com/user-attachments/assets/1365d36a-96ab-4b46-b59a-1c345377a622

cyperdark:
ILOVEJAVASCRIPT

image
image

KotRikD (approved):

PR #214, opened 2024-11-04
fix: Fix building locally on Windows

KapiWilq:
When trying to build tosu locally on Windows (as specified in https://github.com/tosuapp/tosu/blob/master/DEVELOPMENT.md), the process will throw an error when copying the `assets` folder, as there is no `cp` command on Windows.

Trying to compile `master` (https://github.com/tosuapp/tosu/commit/ebf1f2e78faeda04c186247ef6fd1d0f5aa47ad2) locally on Windows:
image

This PR splits the `prepare-htmls` package script into two platform-specific ones.


cyperdark:
If I'm not wrong you need to have those installed on windows

https://gnuwin32.sourceforge.net/packages/coreutils.htm

KotRikD:
use git bash :L

PR #243, opened 2024-11-17
fix: fix unnecessary tournament chat reset

cyperdark:


packages/tosu/src/memory/types.ts
```diff
@@ -162,7 +162,7 @@ export type ITourney =
     | string
     | Error;
 
-export type ITourneyChat = ITourneyManagerChatItem[] | Error;
+export type ITourneyChat = ITourneyManagerChatItem[] | Error | string;
```
KotRikD, line 165: | boolean

packages/tosu/src/memory/stable.ts
```diff
@@ -1024,7 +1024,7 @@ export class StableMemory extends AbstractMemory<OsuPatternData> {
                 }
             }
 
-            return [];
+            return 'skip';
```
KotRikD, line 1027: return false

packages/tosu/src/states/tourney.ts
```diff
@@ -75,6 +75,7 @@ export class TourneyManager extends AbstractState {
 
             const messages = this.game.memory.tourneyChat(this.messages);
             if (messages instanceof Error) throw Error;
+            if (typeof messages === 'string') return;
```
KotRikD, line 78: !messages

KotRikD (changes requested):

KotRikD (approved):

PR #290, opened 2025-01-29
Fix: tosu throwing `ENOENT` errors on paths with trailing spaces inside.

GabuTheDev:
Related https://discord.com/channels/1056534107330445362/1334211010390655016

PR #292, opened 2025-02-01
Fix: Ensure `utf-8` encoding & path cleanup.

GabuTheDev:
The function `cleanPath` sanitizes the paths passed, removing special chars (that would never be in a path, unless incorrect readout), trimming trailling spaces (similar to #290 ) and using utf-8 as encoding (hopefully fixing [this issue](https://discord.com/channels/1056534107330445362/1334946353096425566))

### DIFF
--- a/packages/tosu/src/states/beatmap.ts
+++ b/packages/tosu/src/states/beatmap.ts
@@ -466,12 +466,13 @@ export class BeatmapPP extends AbstractState {
                 const { bpm, bpmMin, bpmMax } = this.lazerBeatmap;
 
                 if (
-                    this.lazerBeatmap.events.backgroundPath !==
+                    cleanPath(this.lazerBeatmap.events.backgroundPath || '') !==
                         menu.backgroundFilename &&
                     !lazerByPass
                 ) {
-                    menu.backgroundFilename =
-                        this.lazerBeatmap.events.backgroundPath || '';
+                    menu.backgroundFilename = cleanPath(
+                        this.lazerBeatmap.events.backgroundPath || ''
+                    );
                 }
 
                 this.previewtime = this.lazerBeatmap.general.previewTime;
--- a/packages/tosu/src/states/beatmap.ts
+++ b/packages/tosu/src/states/beatmap.ts
@@ -315,9 +315,9 @@ export class BeatmapPP extends AbstractState {
                     `beatmapPP updateMapMetadata`,
                     `Skip osu! music theme file`,
                     {
-                        SongsFolder: cleanPath(global.songsFolder),
-                        Folder: cleanPath(menu.folder),
-                        Path: cleanPath(menu.filename)
+                        SongsFolder: global.songsFolder,
+                        Folder: menu.folder,
+                        Path: menu.filename
                     }
                 );
                 return;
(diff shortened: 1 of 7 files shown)


packages/tosu/src/states/beatmap.ts
```diff
@@ -331,18 +330,18 @@ export class BeatmapPP extends AbstractState {
                     `beatmapPP updateMapMetadata`,
                     `Skip new map creation`,
                     {
-                        SongsFolder: global.songsFolder,
-                        Folder: menu.folder,
-                        Path: menu.filename
+                        SongsFolder: cleanPath(global.songsFolder),
```
KotRikD, line 333: Didn't you clean it in osuInstance/lazerInstance ?

KotRikD (changes requested):

packages/tosu/src/utils/converters.ts
```diff
@@ -76,3 +76,19 @@ export const numberFromDecimal = (
...
 };
+
+/**
+ * Joins multiple paths into a single path, replacing invalid Windows path characters in the process
+ *
+ * @param {...string[]} paths Paths to join
+ * @returns {string} Joined path
+ */
+export const cleanPath = (...paths: string[]): string => {
+    paths.map((path) =>
```
KotRikD, line 87: Does this even do something?

like isn't it should be

`paths = paths.map(...)` ?

packages/tosu/src/utils/converters.ts
```diff
@@ -76,3 +76,19 @@ export const numberFromDecimal = (
 
     return value;
 };
+
+/**
+ * Joins multiple paths into a single path, replacing invalid Windows path characters in the process
```
cyperdark, line 81: what about linux? some characters are allowed there

packages/tosu/src/utils/converters.ts
GabuTheDev: > paths = paths.map(...) ?

Oops.

GabuTheDev: Fixed in 172216d15afc160c0441da6cc529a49755abba34

packages/tosu/src/utils/converters.ts
GabuTheDev: Fixed in https://github.com/tosuapp/tosu/pull/292/commits/f67dc2099f6dee51c1eb2100ab2ae65944cf201e

packages/tosu/src/states/beatmap.ts
KotRikD: I guess, it's not good idea. Single place - single responsobility

KotRikD (changes requested):

packages/tosu/src/states/beatmap.ts
GabuTheDev: Fixed at https://github.com/tosuapp/tosu/pull/292/commits/8d82da2e0b59b7e3e45e4a70cbbda9471d00a663

GabuTheDev:
I've ensured that almost all paths are sanitized when new values are appended to the variables. The previous commits mainly addressed the utf-8 error. Since the issue OP mentioned that the background is still not parsed correctly, I've now sanitized those paths as well.

KotRikD (approved):

PR #299, opened 2025-02-22
refactor: Settings Tab UI

GabuTheDev:
This PR overhauls the looks of the settings tab UI inside the dashboard, as well as implementing 6 new web components (7 if you include the save button, which already existed under a different name).

Some mentions:
- This PR also includes a change that has nothing to do with the overhaul of the UI but felt related and it's an QoL (see https://github.com/tosuapp/tosu/commit/c4a49ca9e68bf3968c20fab806fa262df6950b2e)
- Regex is generated. I am not sure if it detects anything else except what I asked for. (I am not a regex guy)
- As per [ck's request](https://i.imgur.com/7aeDkRV.png), the textarea in the new component uses an experimental css property that might not be available to all.

**A detailed video showcase can be found [here](https://youtu.be/hR6yNmB49o4).**

## Overall Preview
 |master | this PR |
 | - | - |
 | Screenshot 2025-02-22 210407 | image |


GabuTheDev:
Tested only on Windows `Chrome 133.0.6943.127` and Linux `Chrome 133.0.6943.98`

packages/server/utils/index.ts
```diff
@@ -114,7 +114,7 @@ export function isRequestAllowed(req: http.IncomingMessage) {
 function isAllowedIP(url: string | undefined) {
     if (!url) return false;
 
-    const allowedIPs = config.allowedIPs.split(',');
+    const allowedIPs = [config.serverIP, ...config.allowedIPs.split(',')];
```
cyperdark, line 117: why, just push `config.serverIP` to allowedIPs

```ts
const allowedIPs = config.allowedIPs.split(',');
allowedIPs.push(config.serverIP);
```

GabuTheDev:
I noticed this PR somehow breaks handling of some buttons? Not sure why tho. Will investigate.

cyperdark (approved):

PR #306, opened 2025-03-15
Fix tosu on linux

KotRikD:


packages/tsprocess/src/process.ts
NelloKudo, line 1: The main reason we support installing tosu in the script using the Windows version is exactly for this reason: we don't expect tosu to be on-par with everything we do upstream. 

Most of our fixes are just to make osu! (and related) work, all while keeping our implementation of tosu working. 
This fix is fine, too, but in the future just point people at using our implementation or using `TOSU_OSU_PATH` for their custom scripts, it'll be way easier for you.

KotRikD: Yes, I understand, that's why it's a joke that I have to crutch for a particular script because most people use it, instead of having a single more or less universal implementation

KotRikD: And using windows build instead of linux one is good too. But still exist people who want native build of tosu instead of calling through API translator as wine do

PR #322, opened 2025-04-17
feat: Disallow tosu startup under OneDrive default directories.

GabuTheDev:
Became a common issue...

The only issue here is that OneDrive users can change the path to whatever they desire, so... this only checks default OneDrive directories. But I guess it's a good addition no matter what.

image


BunnyTub:
This pull request is very funny to me

PR #323, opened 2025-04-18
Update onedrive checking method & Warn to extract tosu from archive

cyperdark:


packages/tosu/src/index.ts
```diff
@@ -44,34 +43,27 @@ const currentVersion = require(process.cwd() + '/_version.js');
...
-        const oneDrivePaths = [
-            'OneDrive',
-            'OneDrive - Personal',
-            'OneDrive - Business',
-            'OneDrive for Business'
-        ].map((path) => resolve(homeDir, path).toLowerCase());
+        const currentPath = getProgramPath().toLowerCase();
+        if (currentPath.includes('temp') && currentPath.includes('appdata')) {
+            wLogger.warn(
+                'Do not run tosu in temp folder. Incase if you running it from archive, please extract tosu from archive (aka .zip)'
```
GabuTheDev, line 50: This is a lot. Just tell them that they shouldn't run it in the archive.

cyperdark: what if they running it not from archive? and it's just a temp folder

GabuTheDev: That would just work

KotRikD (approved):

PR #329, opened 2025-04-21
chore(icon): Cleaned up artifact and use standardized sizes.

GabuTheDev:
Fixed a red artifact (probably appeared due to incorrect conversion, I am assuming the png/jpeg was just renamed to ico, producing it) and replaced the `56x56` to a standardized `64x64` size.

| master | this PR |
|--------|--------|
| old-icon | new-icon |


GabuTheDev:
Just as a notice, Windows tends to cache icons, so the change won't be visible until the next system reboot or by moving the exe into another folder.

cyperdark:
oh yeah, forgot to fix it

PR #335, opened 2025-04-28
feat: attempt to fix ssl

KotRikD:


MilPowder:
\Downloads\tosu-windows-3ca2056\game-overlay\tosu-gameoverlay.zip
Cannot open the file as [zip] archive
Is not archive

PR #341, opened 2025-05-03
feat: Implement `maxAchievable`.

GabuTheDev:
This PR implements a new pp value called `maxAchievable` that indicates the maximum pp value a user could be getting from their play if they fc the rest of the map. This is similar to gosu's old `maxThisPlay` property, so it's a step towards gosu-tosu compatibility. (more or less...)

**Q: How is this any different from `play.pp.fc`?**
**A:** The `play.pp.fc` does not take account of misses; it shows the PP value that could have been achieved only if the player would have **never missed.** The `maxAchievable` value takes account of **all** hit judgements (300s,100s,50s & misses).

---
**!!** There's a few things to consider before merging, such as compatibility issues.
Since this PR also renames `maxAchievedThisPlay` to `maxAchieved`, there are two ways of approaching future issues:
1. Announce the future deprecation of `maxAchievedThisPlay`, but keep both values (old name and new name) even though they provide the same value. (Assuming deprecation comes later, so people have time to switch)
2. Deprecate `maxAchievedThisPlay` (announcing breaking changes) and update all counters available in the tosu counters repository with the new property name (which can be done very easily by me), but the issue still remains with the local/private pp counters that use this

> [!IMPORTANT]
> Either way, the socket will need to be updated with the new changes, which will be a future PR after this gets merged. 

--- 

> [!NOTE]
> The new value has only been tested on osu!std so far, I am not aware of the possible issues with the other rulesets.

### DIFF
--- a/packages/tosu/src/api/types/v1.ts
+++ b/packages/tosu/src/api/types/v1.ts
@@ -96,7 +96,7 @@ interface PP {
 interface GameplayPP {
     current: number;
     fc: number;
-    maxThisPlay: number;
+    maxAchieved: number;
 }
 
 export interface LeaderboardPlayer {
--- a/packages/tosu/src/api/types/v1.ts
+++ b/packages/tosu/src/api/types/v1.ts
@@ -96,7 +96,7 @@ interface PP {
 interface GameplayPP {
     current: number;
     fc: number;
-    maxAchieved: number;
+    maxThisPlay: number;
 }
 
 export interface LeaderboardPlayer {
--- a/packages/tosu/src/api/utils/buildResult.ts
+++ b/packages/tosu/src/api/utils/buildResult.ts
@@ -196,7 +196,7 @@ export const buildResult = (instanceManager: InstanceManager): ApiAnswer => {
             pp: {
                 current: fixDecimals(beatmapPP.currAttributes.pp),
                 fc: fixDecimals(beatmapPP.currAttributes.fcPP),
-                maxAchieved: fixDecimals(beatmapPP.currAttributes.maxAchieved)
+                maxThisPlay: fixDecimals(beatmapPP.currAttributes.maxAchieved)
             },
             keyOverlay: {
                 k1: {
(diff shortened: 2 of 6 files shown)


GabuTheDev:
There are two flaws in this implementation:
1. Rewinding replays in the lazer client breaks the calcs. One fix here would be to just set `maxAchievable` as the replay's pp since we already know how much the play was worth (which would cancel the calcs).
2. For lazer scores, this calc assumes max `osuSmallTickHits` & `osuLargeTickHits` since there's no way of actually predicting them.

packages/tosu/src/api/types/v1.ts
```diff
@@ -96,7 +96,7 @@ interface PP {
 interface GameplayPP {
     current: number;
     fc: number;
-    maxThisPlay: number;
```
KotRikD, line 99: no, v1 api shouldn't be renamed. It's just legacy and it should be 1-to-1 for gosu API

packages/tosu/src/api/utils/buildResult.ts
```diff
@@ -196,7 +196,7 @@ export const buildResult = (instanceManager: InstanceManager): ApiAnswer => {
             pp: {
                 current: fixDecimals(beatmapPP.currAttributes.pp),
                 fc: fixDecimals(beatmapPP.currAttributes.fcPP),
-                maxThisPlay: fixDecimals(beatmapPP.currAttributes.maxThisPlayPP)
```
KotRikD, line 199: same

KotRikD (changes requested):

packages/tosu/src/api/types/v1.ts
GabuTheDev: yeah but iirc, in gosu, `maxThisPlay` was `beatmapPP.currAttributes.maxAchievable` (what this PR is implementing), not `beatmapPP.currAttributes.maxAchieved`

KotRikD: Yep, we should support Legacy API as much as we can, so yeah

KotRikD (approved):

PR #350, opened 2025-05-15
feat: ingame overlay editor

storycraft:
1. Hard coded keybind from the previous overlay(shift + ctrl + space)
2. Electron offscreen input is somewhat limited. I did best effort.
3. IME doesn't work correctly yet if osu! is not using IME(not focused on text input). However this is not a big problem for configuration. English input works without IME.
4. Cursor works

Showcase: https://www.youtube.com/watch?v=1yhLjEBmb4Q

### DIFF
--- a/packages/ingame-overlay/electron.vite.config.mjs
+++ b/packages/ingame-overlay/electron.vite.config.mjs
@@ -38,6 +38,5 @@ export default defineConfig({
       outDir: 'dist/preload',
       minify: true,
     },
-    plugins: [bytecodePlugin()],
   },
 });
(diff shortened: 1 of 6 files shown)


KotRikD (dismissed):

storycraft:
There was an issue pressing Alt or F10 freezing window while ingame configuration is enabled. I updated package with fix

storycraft:
Also while working on this, I improved opengl overlay rendering performance with fixes and it will not cause freezing on some devices anymore.

KotRikD (approved):

PR #358, opened 2025-05-23
fix: pp calc improvements

### LINKED ISSUE #351: Passing a floating-point negative number to rosu causes errors.
The `updateStarsAndPerformance` functions throws an error due to `invalid type`. Error passed via wasm (rosu).
We're passing a negative float number through one of the properties of a performance calculator func or rosu returns a negative value through wasm, which expects an u32.

There are more chances of being the later one, because the user reported that the value decreases over time (meaning it is accumulating negatively lol).

Image

Repoted on discord by `matthieu103` | [message link](https://discord.com/channels/1056534107330445362/1185957776665628764/1372637336558506094)

GabuTheDev:
This PR fixes some issues regarding performance points calculation, specifically:

- [x] [Fixes noticeable pp fc calc performance degradation for the Mania ruleset.](https://discord.com/channels/1056534107330445362/1185957776665628764/1375229211018657983)
- [x] Fixes pp calc for FC (now the value should only decrease from it's maximum value)
- [x] Changes pp value assignment order. (if one fails, it shouldn't stop the others from updating)
- [x] Fixes #351 
- [x] Some refactoring.. nothing crazy

tested on both lazer and stable. (I've learned my lesson..)

### DIFF
--- a/packages/tosu/src/states/gameplay.ts
+++ b/packages/tosu/src/states/gameplay.ts
@@ -1,4 +1,4 @@
-import rosu from '@kotrikd/rosu-pp';
+import rosu, { PerformanceArgs } from '@kotrikd/rosu-pp';
 import { ClientType, config, wLogger } from '@tosu/common';
 
 import { AbstractInstance } from '@/instances';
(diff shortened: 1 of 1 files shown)


packages/tosu/src/states/gameplay.ts
```diff
@@ -508,39 +508,44 @@ export class Gameplay extends AbstractState {
...
             ) {
-                if (this.gradualPerformance) this.gradualPerformance.free();
-                if (this.performanceAttributes)
-                    this.performanceAttributes.free();
-
-                const difficulty = new rosu.Difficulty({
-                    mods: removeDebuffMods(this.mods.array),
-                    lazer: this.game.client === ClientType.lazer
-                });
+                [this.gradualPerformance, this.performanceAttributes].forEach(
```
cyperdark, line 522: dont do that

packages/tosu/src/states/gameplay.ts
```diff
@@ -508,39 +508,44 @@ export class Gameplay extends AbstractState {
             const currentState = `${menu.checksum}:${menu.gamemode}:${this.mods.checksum}:${menu.mp3Length}`;
             const isUpdate = this.previousState !== currentState;
 
+            const commonOptions = {
```
cyperdark, line 511: commonParams

packages/tosu/src/states/gameplay.ts
```diff
@@ -508,39 +508,44 @@ export class Gameplay extends AbstractState {
...
-                }).calculate(currentBeatmap);
+                this.performanceAttributes = new rosu.Performance(
+                    commonOptions
+                ).calculate(currentBeatmap);
 
                 this.previousState = currentState;
             }
 
             if (!this.gradualPerformance || !this.performanceAttributes) {
+                const missingComponents = {
```
cyperdark, line 540: do not create new object for that

packages/tosu/src/states/gameplay.ts
```diff
@@ -508,39 +508,44 @@ export class Gameplay extends AbstractState {
...
+                    gradualPerformance: !this.gradualPerformance,
+                    performanceAttributes: !this.performanceAttributes
+                };
                 wLogger.debug(
                     ClientType[this.game.client],
                     this.game.pid,
-                    `gameplay updateStarsAndPerformance One of the things not ready`,
-                    `gradual: ${this.gradualPerformance === undefined} - attributes: ${this.performanceAttributes === undefined}`
+                    'gameplay updateStarsAndPerformance: Missing required components',
+                    JSON.stringify(missingComponents)
```
cyperdark, line 548: dont

packages/tosu/src/states/gameplay.ts
```diff
@@ -558,67 +563,79 @@ export class Gameplay extends AbstractState {
...
             if (fcPerformance) {
                 beatmapPP.currAttributes.fcPP = fcPerformance.pp;
                 beatmapPP.updatePPAttributes('fc', fcPerformance);
             }
 
+            const maxAchievablePerformance = new rosu.Performance({
+                ...calcOptions,
+                misses: this.hitMiss
+            }).calculate(this.performanceAttributes);
+
```
cyperdark, line 638: why you are doing `...calcOptions,` when you can set `calcOptions.misses = 0;` and pass calcOptions as it is, and update `calcOptions.misses = this.hitMiss;`

instead of making copy of all objects twice

cyperdark (changes requested):

packages/tosu/src/states/gameplay.ts
GabuTheDev: any reasons?

packages/tosu/src/states/gameplay.ts
GabuTheDev: This is ensuring `calcOptions` is not being modified throught the calcs.
While you're right, it creates copies, I believe this is more of a preference instead of a "performance" issue, if that's why you pointed this out.

packages/tosu/src/states/gameplay.ts
cyperdark: why you need foreach on two simple conditions?

cyperdark (approved):

PR #360, opened 2025-05-25
fix: in-game overlay returning null code.

GabuTheDev:
Usually, when a process child returns null as exit code, it means that the process was not terminated normally (Likely terminated by a signal instead). This PR just adds the signal (if any) to the logging part. (a step to easily debug the few reports we've got so far with `unknown code: null`) 

More info about the signals [here](https://nodejs.org/api/process.html#signal-events)


GabuTheDev:
This will likely still return null in most cases as WIndows does not report as many signals as linux does, but whatever.

PR #364, opened 2025-05-30
feat: redirect in-game overlay logging, process status tracking

storycraft:
* Redirect ingame overlay stdout, stderr to debug, debugError (they are verbose)
* Properly track the status of overlay process. This fix the issue tosu needs to restart completely to start ingame overlay again once the overlay is closed by user.
* close #360 pr(I included the commit before the refactor)
* add debug log when ingame overlay fail to launch due to an another instance already running(by another tosu or manual launch).
* a bit of asyncification (async `setTimeout`)
* fix ingame overlay failing due to lack of ipc access permission when osu! is launched with lower permission user than tosu
* apply start option to prefer discrete gpu

### DIFF
--- a/packages/tosu/src/instances/manager.ts
+++ b/packages/tosu/src/instances/manager.ts
@@ -145,14 +145,14 @@ export class InstanceManager {
     }
 
     async runWatcher() {
-        for (;;) {
+        while (true) {
             await this.handleProcesses();
             await setTimeout(1000);
         }
     }
 
     async runDetemination() {
-        for (;;) {
+        while (true) {
             if (!this.platformType) {
                 const platform = platformResolver(process.platform);
                 this.platformType = platform.type;
--- a/packages/tosu/src/instances/manager.ts
+++ b/packages/tosu/src/instances/manager.ts
@@ -7,6 +7,8 @@ import {
     wLogger
 } from '@tosu/common';
 import { runOverlay } from '@tosu/ingame-overlay-updater';
+import { ChildProcess } from 'node:child_process';
+import { setTimeout } from 'node:timers/promises';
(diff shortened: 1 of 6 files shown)


packages/tosu/src/instances/manager.ts
```diff
@@ -134,47 +133,73 @@ export class InstanceManager {
...
+        }
     }
 
-    runDetemination() {
-        if (!this.platformType) {
-            const platform = platformResolver(process.platform);
-            this.platformType = platform.type;
-        }
+    async runDetemination() {
+        for (;;) {
```
KotRikD, line 155: same for this what?

KotRikD:
Well I don't like that construction
```js
for (;;)
```

I understand that this is `while (true)`. So my question why not using it either?

KotRikD (changes requested):
lookup

storycraft:
> Well I don't like that construction
> 
> ```js
> for (;;)
> ```
> 
> I understand that this is `while (true)`. So my question why not using it either?

Imo `for(;;)` fits better than `while (true)` for *unconditional* loop, `while (true)` still have condition `true`. It's preference, so let me change it if you don't like.

KotRikD:
> > Well I don't like that construction
> > ```js
> > for (;;)
> > ```
> > I understand that this is `while (true)`. So my question why not using it either?
> 
> Imo `for(;;)` fits better than `while (true)` for _unconditional_ loop, `while (true)` still have condition `true`. It's preference, so let me change it if you don't like.

Well yeah, good point, but for now lets use `while(true)`. I will think about swapping to `for(;;)`

KotRikD (approved):

PR #365, opened 2025-05-31
fix: ingame overlay cursor

storycraft:
* remove obsolete cursor images
* remove obsolete cursor image div hack
* fix default cursor style


cyperdark:
@storycraft did you test it?

storycraft:
> @storycraft did you test it?

Yes. I actually tested it.
* The old cursor div hack was invisible(had zero opacity) in both ingame overlay configuration and dashboard configuration.
* The actual cursor was not being restored properly once the cursor changed because `arrow` is not valid css cursor value. This didn't happen to old `custom-cursor` because it used `arrow` cursor image file. Fixing to `default` restores it to arrow cursor.

PR #367, opened 2025-06-06
Fix mania freezes & Fix pp related issues

GabuTheDev:
at this point, my PRs should be tagged the `bug` label.
or.. we should add test scenes :D (haha..)

### DIFF
--- a/packages/tosu/src/states/beatmap.ts
+++ b/packages/tosu/src/states/beatmap.ts
@@ -412,6 +412,11 @@ export class BeatmapPP extends AbstractState {
                 `[${totalTime}ms] Spend on opening beatmap`
             );
 
+            const commonParams = {
+                mods: removeDebuffMods(currentMods.array),
+                lazer: this.game.client === ClientType.lazer
+            };
+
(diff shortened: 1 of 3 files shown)


cyperdark:
> we should add test scenes :D (haha..)

already has but only for cherry to test things

GabuTheDev:
> already has but only for cherry to test things

well... I guess one more thing that could go in my to-do for this project.
But it's gonna take a while cuz like.. what tests can I do for tosu..?

GabuTheDev:
So i found two issues here:

1. the combo of an fc play can be sometimes lower than the max achievable combo on that map due to missed slider ends (or so I think).
This can be fixed only if we have a way of actually tracking the slider end misses.

2. The 300s hit value on the current pp calc is always lower than the predicted fc value.
The current way of predicting it is to subtract each hit judgment that isn't a 300 from the max possible 300s but that seems to be wrong as it can sometimes different up to 400 hits (which is insane)

KotRikD (approved):

Issue #361, opened 2025-05-26
In Game Overlay freeze issue


kenz2874:
When i try enable in game overlay it freeze when i alt+tab

cyperdark:
# ` 📣 ` How to report an issue
ㅤ

1. Enable __**debug_log**__ in tosu settings
2. Describe issues as detailed as possible
3. Attach latest file from **logs/** folder 
4. Attach screenshot or video of problem
5. Attach tsosu.env
6. if you recording a video __**make sure to place tosu**__ console on top of osu, so we can see when issue happens
7. What **OS** do you use? (e.g., `Windows 10` or `Ubuntu 22.04.5`)
8. In which client does error happened? (e.g., `stable`, `cutting-edge`, `lazer`)

GabuTheDev:
Hello @kenz2874! This issue got fixed in the latest release!
If you're still having issues, please open a new issue thread.

PR #374, opened 2025-06-09
feat: ingame overlay update

### LINKED ISSUE #356: Disabling the in-game overlay setting does not end the overlay process.
I believe the overlay exe should be terminated if the setting is turned off and the overlay is running already. Right now, disabling it requires a restart, which is not intuitive at all.v

storycraft:
* Fix #356
* Maybe #361
* Update overlay library

## Overlay library changelog
* Overlay DLL is now signed(v0.6.5)
* Improve opengl overlay interoperability with other overlay software(yes afterburner and rtss, I am looking at you)
* Fix alt+tab freezing on AMD gpu
* Overlay is not lost on swapchain recreation


storycraft:
Was thinking including another fix here but will be on next pr

cyperdark (approved):

PR #376, opened 2025-06-13
feat: Add decorator to measure time

cyperdark:


packages/common/utils/logger.ts
```diff
@@ -76,3 +86,46 @@ function writeLog(type: string, ...args: any[]) {
...
+
+        const t1 = performance.now();
+        const result = originalMethod.apply(this, args);
+        const time = performance.now() - t1;
+
+        if (time >= 1 && client) {
+            const coloredText = colorText('timing', 'timing');
+            console.log(
+                coloredText,
+                `[${client}]`,
```
GabuTheDev, line 115: should match the rest of the logging functions (they do not have square brackets around client name)
```suggestion
                `${client}`,
```

packages/tosu/src/states/gameplay.ts
```diff
@@ -281,6 +282,7 @@ export class Gameplay extends AbstractState {
         }
     }
 
+    @measureTime
     updateKeyOverlay() {
         try {
             const result = this.game.memory.keyOverlay(this.mode);
```
GabuTheDev, line 288: hmm

https://github.com/user-attachments/assets/0e645214-f0db-4f51-a376-f1b55259bd48

packages/common/utils/logger.ts
```diff
@@ -76,3 +86,46 @@ function writeLog(type: string, ...args: any[]) {
...
+                (this as any).game.pid,
+                `${target.constructor.name}.${propertyKey} [${time.toFixed(2)}ms] Total spent time`
+            );
+        } else if (time >= 1) {
+            const coloredText = colorText('timing', 'timing');
+            console.log(
+                coloredText,
+                `${target.constructor.name}.${propertyKey} [${time.toFixed(2)}ms] Total spent time`
```
GabuTheDev, line 123: do we really have to mention that it's the `total spent time` each log?
can't we just.. `${target.constructor.name}.${propertyKey} executed in ${time.toFixed(2)}ms`
or simply `${target.constructor.name}.${propertyKey} [${time.toFixed(2)}ms]`

cyperdark: leftover from previous approach i was trying to do

KotRikD (approved):

GabuTheDev (approved):
lgtm

PR #384, opened 2025-06-15
feat: Sort local/external counters by name

GabuTheDev:
makes more sense than the current sorting.


cyperdark:
no, sorting by name is bad idea

cyperdark:
updated default order for overlays, check it and if you want to move some let me know in dms

PR #393, opened 2025-06-21
fix: Fix cleanup after update when updating from dashboard

### LINKED ISSUE #366: tosu cannot do update cleanup if updated via the dashboard.
When updating via the dashboard (auto updating is off), tosu fails to delete `tosu_old.exe` and the archive.
It also makes the new tosu run before the old one closes, so it always starts with "port already in use"

cyperdark:
closes #366

### DIFF
--- a/packages/server/router/index.ts
+++ b/packages/server/router/index.ts
@@ -336,15 +336,9 @@ export default function buildBaseApi(server: Server) {
         }
     );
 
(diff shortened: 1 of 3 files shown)


GabuTheDev (approved):
Seems fine.
Can't really test it without having an actual update.

KotRikD (approved):

PR #408, opened 2025-07-13
feat: New API endpoint to generate debug configuration report

storycraft:
* See #398


cyperdark:
rewrote page itself with some fixes

127 0 0 1_24050_api_generateReport

PR #427, opened 2025-08-17
feat: HIde deleted overlays & show them back after redownload

### LINKED ISSUE #390: Removing local counters via the dashboard does not remove them from the ingame overlay.
they are still present on the game overlay (and working!)
it is worse when I reload overlays

Image

cyperdark:
closes #390


packages/server/utils/counters.ts
```diff
@@ -162,10 +162,7 @@ function rebuildJSON({
                 const hasQuery = queryArr.some(
                     (q) =>
                         item.name.toLowerCase().includes(q) ||
-                        item.author.toLowerCase().includes(q) ||
```
GabuTheDev, line 165: You removed querying by `compatibleWith`. Is this intended?

cyperdark: yes, otherwise searching tosu return all counters instead of giving only tosu debug

GabuTheDev (approved):

KotRikD (approved):

PR #442, opened 2025-08-24
fix: ensure paths are sanitized correctly.

GabuTheDev:
- function now actually does something (blame it on me, I'm the culprit)
- changed it's name to actually represent what it does.
- it's now using the system's path separator and depends on nodejs to normalize it to utf-8

things I have to do bef merge:
- [x] add back jsdoc


storycraft:
Do we actually need sanitizing? If the path is invalid, it will error on reading anyways. Even If the path is sanitized, files or folders pointed by path would become invalid.

GabuTheDev:
> Do we actually need sanitizing?

There are beatmaps that have a trailing space in their folder name, thus making nodejs unable to read the folder. Sanitising the folder name before joining the path altogether fixes this. Is this scenario not enough of a reason?

PR #469, opened 2025-09-19
fix: defer config file watching until init.

storycraft:
* Initial load of config file first and run watcher after initialization is finished.
* Use `fs.watch` instead of `setTimeout`.

### DIFF
--- a/packages/common/utils/config.ts
+++ b/packages/common/utils/config.ts
@@ -211,28 +211,23 @@ export const updateConfigFile = async () => {
     }
 };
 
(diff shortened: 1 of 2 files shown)


storycraft:
This also ensure overlay to not start when updating. Like #467, but different approach.

cyperdark:
closing due to https://github.com/tosuapp/tosu/pull/473 PR

cyperdark:
closing due to https://github.com/tosuapp/tosu/pull/473 PR

Issue #477, opened 2025-09-26
tosu does not work for the tourney client on linux platforms.


GabuTheDev:
Originally reported here: https://discord.com/channels/1056534107330445362/1421188625340567622

As far as my investigation went, tosu does find the tourney player clients, but does not categorize them as such because the `--spectateclient` argument does exist on the process, but `tsprocess.node` does not read it.

 
<img width="881" height="430" alt="Image" src="https://github.com/user-attachments/assets/5de6806b-07fb-462a-add3-d7a6e20c3d1e" />

<img width="481" height="167" alt="Image" src="https://github.com/user-attachments/assets/d6c7a581-9b0a-4b02-b7bd-7edf0c980a8f" />

https://github.com/tosuapp/tosu/blob/b29f60dd789a4b1ce2db2e7149459c299f644497/packages/tosu/src/instances/manager.ts#L119-L124

https://github.com/tosuapp/tosu/blob/b29f60dd789a4b1ce2db2e7149459c299f644497/packages/tsprocess/src/process.ts#L58-L66

GabuTheDev:
cc @KotRikD @xxCherry (I guess)

KotRikD:
~~https://github.com/tosuapp/tosu?tab=readme-ov-file#linux-notice~~

KotRikD:
will check

PR #478, opened 2025-09-27
fix: tournament on linux

KotRikD:


KotRikD:
@xxCherry please look additionally

KotRikD:
closes #477

PR #480, opened 2025-09-27
chore: Update osu!(lazer) mods' settings

KapiWilq:
- Adds "[Also fail when missing a slider tail](https://github.com/ppy/osu/blob/20dbe096e03e043143388eab62e1650a3be1ea2a/osu.Game.Rulesets.Osu/Mods/OsuModSuddenDeath.cs#L22-L23)" in osu!standard's "Sudden Death" and "[Require perfect hits](https://github.com/ppy/osu/blob/1e61abde7498e7d101dfa81b149ddc4d49d3fa36/osu.Game.Rulesets.Mania/Mods/ManiaModPerfect.cs#L14-L15)" in osu!mania's "Perfect"
- Fixes "Style" in osu!standard's "[Approach Different](https://github.com/MaxOhn/rosu-mods/blob/6c5634488a076b35a1a0e8e9de6e7fefc7b05985/src/generated_mods.rs#L373-L379)" and "Direction" in both osu!standard's "[Barrel Roll](https://github.com/MaxOhn/rosu-mods/blob/6c5634488a076b35a1a0e8e9de6e7fefc7b05985/src/generated_mods.rs#L360-L366)" and osu!mania's "[Cover](https://github.com/MaxOhn/rosu-mods/blob/6c5634488a076b35a1a0e8e9de6e7fefc7b05985/src/generated_mods.rs#L1097-L1103)" being the wrong type

---

NOTE: `rosu-mods` (and therefore `rosu-pp` by extension) are "[not aware](https://github.com/MaxOhn/rosu-mods/blob/6c5634488a076b35a1a0e8e9de6e7fefc7b05985/src/generated_mods.rs#L1053-L1057)" of the "Require perfect hits" setting, so they think it's not valid. This breaks various fields that rosu calculates/provides when the setting is enabled. So my question is: should I comment out the memory reading part or leave it as is?

https://github.com/user-attachments/assets/18acf54f-0df7-4e01-a101-2d64c82e81fa

### DIFF
--- a/packages/tosu/src/memory/lazer.ts
+++ b/packages/tosu/src/memory/lazer.ts
@@ -2194,25 +2194,49 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
                 break;
             }
             case 'SD': {
+                const settings: any = {};
                 const restartBindable = this.process.readIntPtr(
                     modObject + 0x10
                 );
 
-                mod.settings = {
-                    restart: this.process.readByte(restartBindable + 0x40) === 1
-                };
+                const failOnSliderTailBindable = this.process.readIntPtr(
+                    modObject + 0x20
+                );
 
+                if (this.selectedGamemode === 0) {
+                    settings.fail_on_slider_tail =
+                        this.process.readByte(
+                            failOnSliderTailBindable + 0x40
+                        ) === 1;
+                }
+
+                settings.restart =
+                    this.process.readByte(restartBindable + 0x40) === 1;
+
+                mod.settings = settings;
                 break;
             }
             case 'PF': {
+                const settings: any = {};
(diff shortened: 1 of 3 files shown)


packages/tosu/src/memory/lazer.ts
```diff
@@ -2590,7 +2614,7 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
 
                 mod.settings = {
                     scale: this.process.readFloat(scaleBindable + 0x40),
-                    style: this.process.readInt(styleBindable + 0x40)
+                    style: `${this.process.readInt(styleBindable + 0x40)}`
```
KotRikD, line 2617: why it's need to make additional ``

PR #492, opened 2025-10-11
feat: Implement the rest of lazer's settings (except for keybinds)

KapiWilq:
Omitted keybinds because I want to do all of them rather than only the quick retry one, and I haven't convinced myself enough yet to populate ruleset keybinds only during gameplay. Also that would probably require an API structure change if actually implemented due to some actions having more than one possible keybind

### DIFF
--- a/packages/common/enums/osu.ts
+++ b/packages/common/enums/osu.ts
@@ -306,6 +306,7 @@ export enum FrameworkSetting {
     ShowLogOverlay,
 
     AudioDevice,
+    AudioUseExperimentalWasapi,
     VolumeUniversal,
     VolumeEffect,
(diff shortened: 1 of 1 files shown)


KotRikD (approved):

KotRikD:
@KapiWilq We're updated our auto tool for your offsets, if all good with pr, you can merge it

KapiWilq:
[They added a new setting since](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Configuration/FrameworkConfigManager.cs#L83), so I'll commit it shortly. Other than that, LGTM

> if all good with pr, you can merge it

No merge perms, unless this part was not directed at me

PR #513, opened 2025-12-04
feat(linux)!: use XDG data/cache/config paths

K4zoku:
BREAKING CHANGE: The storage paths for data, cache, and configuration files on Linux have been updated to follow the XDG Base Directory Specification.

- Overlay and logs files are now stored in `$XDG_DATA_HOME/tosu` (default: `~/.local/share/tosu`).
- Cache files are now stored in `$XDG_CACHE_HOME/tosu` (default: `~/.cache/tosu`).
- Configuration files are now stored in `$XDG_CONFIG_HOME/tosu` (default: `~/.config/tosu`).

This change improves compliance with Linux standards but may require users to migrate their existing files from the previous storage paths to the new XDG-compliant paths

### DIFF
--- a/packages/common/utils/directories.ts
+++ b/packages/common/utils/directories.ts
@@ -69,9 +69,8 @@ export function getCachePath() {
         const cachePath = path.join(xdgCacheHome, 'tosu');
         ensureDirectoryExists(cachePath);
         return cachePath;
-    } else {
-        return path.join(getProgramPath(), '.cache');
     }
+    return path.join(getProgramPath(), '.cache');
 }
 
 export function getProgramPath() {
@@ -100,9 +99,8 @@ export function getDataPath() {
         const dataPath = path.join(xdgDataHome, 'tosu');
         ensureDirectoryExists(dataPath);
         return dataPath;
-    } else {
-        return getProgramPath();
     }
+    return getProgramPath();
 }
 
 export function getConfigPath() {
@@ -112,7 +110,6 @@ export function getConfigPath() {
         const configPath = path.join(xdgConfigHome, 'tosu');
         ensureDirectoryExists(configPath);
         return configPath;
-    } else {
-        return getProgramPath();
     }
+    return getProgramPath();
 }


packages/common/utils/directories.ts
```diff
@@ -43,20 +50,28 @@ export function recursiveFilesSearch({
...
 
 export function getCachePath() {
-    return path.join(getProgramPath(), '.cache');
+    if (process.platform === 'linux') {
+        const xdgCacheHome =
+            process.env.XDG_CACHE_HOME || path.join(homedir(), '.cache');
+        const cachePath = path.join(xdgCacheHome, 'tosu');
+        ensureDirectoryExists(cachePath);
+        return cachePath;
+    } else {
```
KotRikD: why you need else statement if you close under if guard condition at 71?

packages/common/utils/directories.ts
```diff
@@ -67,13 +82,37 @@ export function getProgramPath() {
...
+
+export function getDataPath() {
+    if (process.platform === 'linux') {
+        const xdgDataHome =
+            process.env.XDG_DATA_HOME ||
+            path.join(homedir(), '.local', 'share');
+        const dataPath = path.join(xdgDataHome, 'tosu');
+        ensureDirectoryExists(dataPath);
+        return dataPath;
+    } else {
```
KotRikD: same

packages/common/utils/directories.ts
```diff
@@ -67,13 +82,37 @@ export function getProgramPath() {
...
+}
+
+export function getConfigPath() {
+    if (process.platform === 'linux') {
+        const xdgConfigHome =
+            process.env.XDG_CONFIG_HOME || path.join(homedir(), '.config');
+        const configPath = path.join(xdgConfigHome, 'tosu');
+        ensureDirectoryExists(configPath);
+        return configPath;
+    } else {
```
KotRikD: same

KotRikD (changes requested):

packages/common/utils/directories.ts
K4zoku: In the first round I was using switch case with entire possible value listed. This one is left over after refactor to if. I will remove the else.

PR #528, opened 2026-01-18
feat: remove build steps for internal packages

### LINKED ISSUE #526: RFC: Transition to TypeScript Path Mapping.
It became quite annoying in tosu development to require manual build step for any tosu package.
When doing changes to, say `@tosu/common`, it shouldn't require a build process to js, because most of the codebase is already in typescript.

My proposal is to switch to TS path mapping, which is documented here: https://www.typescriptlang.org/tsconfig/#paths

storycraft:
* Close #526 
* `prepare` steps are still needed for css generation.
* Remove `build` script from internal packages `@tosu/server`, `@tosu/common` and `@tosu/ingame-overlay-updater`.
* Point `index.ts` in `main` field of `package.json`.
* Remove `rootDir` in `tosu` package`s tsconfig.json to build internal dependencies.
* Fix `pkgAssetsPath` since we don't have `dist` folder now.
* Fix `postBuild.ts` types.

### DIFF
--- a/packages/tosu/src/postBuild.ts
+++ b/packages/tosu/src/postBuild.ts
@@ -1,11 +1,12 @@
 import fs from 'fs';
 import path from 'path';
-import ResEdit from 'resedit';
+import { load } from 'resedit/cjs';
 import semverParse from 'semver/functions/parse';
 
 async function windowsPostBuild(output: string) {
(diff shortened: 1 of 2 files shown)


KotRikD:
Ohhh, why I don't think about that in that case...

KotRikD:
Ah yeah, because vercel's pkg sucks

Issue #527, opened 2026-01-18
broken score reading in tournament mode on cutting edge 20260116


ILW8:
Cutting edge [20260116](https://osu.ppy.sh/home/changelog/cuttingedge/20260116) seems to have broken reading of scores in tournament mode (`tourney.ipcClients[].gameplay.score` is always 0)

KotRikD:
😐

KotRikD:
Found problem, broken offset for gameplay, so this is not only tournament issue, will fix asap

for temp resolution: teamLeftScore + teamRightScore seems to be working

KotRikD:
@ILW8 just in case, https://github.com/tosuapp/tosu/actions/runs/21107448948?pr=529 try this build, should be fixed

ILW8:
> for temp resolution: teamLeftScore + teamRightScore seems to be working

I suggested this and put out a fix for the overlay using left and right, but it got stuck after a couple maps as well (someone else ran into this issue, wasn't a tournament I was running)

KotRikD:
Well waiting for testing, if all works good, I will close issue

btw tested on #roc2025 , seems working but who knows? :D

ILW8:
I have heard back that it seems to work again, thanks for the quick fix.

Closed by #529

PR #529, opened 2026-01-18
fix: allow PNA for chrome more than 142 and fix scoring

KotRikD:
https://chromestatus.com/feature/5152728072060928

+ latest cutting edge brokes score


KotRikD:
fixes #527

PR #532, opened 2026-01-20
feat: Implement ingame overlay profiles & Separate profile selector for obs

cyperdark:
https://github.com/user-attachments/assets/6467b613-f72f-4c47-a412-1c93229c4a6f

### DIFF
--- a/packages/ingame-overlay/renderer/index.html
+++ b/packages/ingame-overlay/renderer/index.html
@@ -34,7 +34,7 @@
                 <span>obs</span>
             </div>
             <div v-if="settings.profiles.length > 0" class="scroll">
-                <div v-for="value in settings.profiles" class="flexer profile-item">
+                <div v-for="(value, index) in settings.profiles" class="flexer profile-item">
                     <span v-text="value.overlays.length"></span>
                     <input type="text" class="input" placeholder="profile name" v-model="value.name">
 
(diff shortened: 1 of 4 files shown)


storycraft (approved):

GabuTheDev:
I don't like the "obs profile selector". I'd much rather have the freedom of choosing any from just using a query url.
There will be some people which will want multiple obs profiles for different scenes they have.

I am also still against merging this PR as the user interface will be replaced soon after.
Instead I would keep this PR around just for the sake of having a real prototype to reference to.
Providing a new service then changing it's behaviour or looks provides a bad experience for the userbase.

cyperdark:
> I don't like the "obs profile selector". I'd much rather have the freedom of choosing any from just using a query url.
There will be some people which will want multiple obs profiles for different scenes they have.

fixed, now most of people can use selector in ingame-overlay and who want more freedom can use `profile` query parameter

<br>

> I am also still against merging this PR as the user interface will be replaced soon after.

like a month or two? also this PR not only have profiles but fixes overall ingame-overlay syncing which caused reverting to old position of overlays and not realtime updates for all ingame-overlays

GabuTheDev:
> like a month or two?

Hopefully before mid March

KotRikD:
@codex review

PR #524, opened 2026-01-12
feat: Switch to `osu-native` pp calculator

cyperdark:
- [x] cleanup
- [ ] cleanup again
- [ ] testing
- [ ] fix converted beatmaps
- [ ] add support for mod settings
- [ ] get ar, od, hp, cs from somewhere
- [ ] fix whatever is this
image

other notes from gabu:
- [ ] The endpoint `/api/calculate/pp` is broken
- [ ] [`hitWindow` undefined](https://github.com/tosuapp/tosu/pull/524/changes#diff-086e35bda98df42528415bafc51d8dd9acdf62ef63a53e3d4469d6844f2ac489R683)
- [ ] [strains still use rosu](https://github.com/tosuapp/tosu/pull/524/changes#diff-879212708ba77293e2638bd6c2971f167e50b2bebb38b50593af91d0c3780503R207)
- [ ] [do not forget about `ppDifficulty`](https://github.com/tosuapp/tosu/pull/524/changes#diff-086e35bda98df42528415bafc51d8dd9acdf62ef63a53e3d4469d6844f2ac489R300)
- [ ] should check for race issues with `this.isCalculating == false`

PR #541, opened 2026-01-30
feat: very initial osu-native support

xxCherry:
this is very quick and dirty, not final at all, might require some adjustments from osu-native side


xxCherry:
work towards this is moved to this pr: https://github.com/tosuapp/tosu/pull/524

PR #549, opened 2026-02-06
fix: Make profile button above overlays

cyperdark:


packages/ingame-overlay/renderer/ingame.css
```diff
@@ -428,6 +428,7 @@ iframe {
     opacity: 1;
 
     transform: translate(-50%, -50%);
+    z-index: 100;
```
KotRikD: 999?

cyperdark: done 👍

KotRikD (approved):

PR #565, opened 2026-02-27
fix: Fix logger formatting

### LINKED ISSUE #564: Logger cannot highlight strings properly.
Was working fine in logs rework (https://github.com/GabuTheDev/tosu/commit/c2fac7ea7409bdbebb0c84a1bea6085aafc29172)
regressed in ck's rewrite (https://github.com/tosuapp/tosu/commit/6bb07c220c8ccc39a9bc3cb9a0f9c21ae8878a6f)

my regex was that long for a reason :/

1. multiple words are not getting highlighted.

Image

2. the char `%` cannot be escaped anymore.

In my commit, doing `This is at %96\% progress%` could property select and highlight `96% progress`.

cyperdark:
closes #564

### DIFF
--- a/packages/common/utils/logger.ts
+++ b/packages/common/utils/logger.ts
@@ -10,6 +10,7 @@ import { ensureDirectoryExists, getDataPath } from './directories';
 import { progressManager } from './progress';
 
 const HighlightColor = '\x1b[1m\x1b[37m';
+const HighlightRegex = /%(.*?)%/g;
 
 let logStream: fs.WriteStream | null = null;
 
@@ -146,7 +147,7 @@ function logFile(type: string, ...args: any[]) {
     const message = args
         .map((arg) => {
             if (typeof arg === 'string') {
-                return arg.replace(/%(\S+)%/g, '$1');
+                return arg.replaceAll(HighlightRegex, '$1');
             }
             if (typeof arg === 'object') {
                 return (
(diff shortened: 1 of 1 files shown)


packages/common/utils/logger.ts
```diff
@@ -10,6 +10,7 @@ import { ensureDirectoryExists, getDataPath } from './directories';
 import { progressManager } from './progress';
 
 const HighlightColor = '\x1b[1m\x1b[37m';
+const HighlightRegex = /%%(.*?)%%/g;
```
GabuTheDev: Why the double percent symbol?
Why would we want to change from `%` to `%%` instead of using the escape symbol as everybody else?

GabuTheDev: > i have no clue what this abomination of regex does

It's just a regex that uses grouping.

`(^|[\s([])%`: looks for a `%` that has a whitespace before it, or the string starts with it.
`((?:\\.|[^%\\])+)` everything in-between two `%`, ignoreing `\%`
`%(?=[.,:;!?\s)\]]|$)` checks for ending with % that has either space after, punctuation, or brackets, so like, basic cases with symbols.
`(\\%)`: this one is to make sure `\%` are escaped.

> first time hearing about this one

ever used `\n`? or `\r`? `\\` ? the `\` before another ascii symbol is to escape them.
https://developer.mozilla.org/en-US/docs/Glossary/Escape_character

PR #590, opened 2026-04-05
feat: Switch to lazer pp calc

storycraft:
* Supersede and closes #524.

## Changes
* [ ] Testing with more ranked scores.
* [x] catch pp calc is broken
* [ ] Replace rosu in `@tosu/server`.
* [x] Lazer mods config
* [x] Bundling natives.
* [x] Perfect hit result calculation for editor. 
* [x] Hit results generation for arbitary acc.
* [x] Gradual difficulty calculation.
* [x] Converted maps.
* [x] Beatmap difficulty(ar, cs, od, hp) calculation with mods applied.
* [x] Strains for each gamemodes.
* [x] Hit results conversion from/to lazer and stable. 
* [x] Correct pp calc on lazer replay rewind.


packages/tosu/src/states/beatmap.ts
```diff
@@ -663,35 +664,32 @@ export class BeatmapPP extends AbstractState {
...
-            switch (strains.mode) {
+            switch (this.beatmap.mode) {
                 case 0:
-                    updateWithOffset('aim', strains.aim);
-                    updateWithOffset('aimNoSliders', strains.aimNoSliders);
-                    updateWithOffset('flashlight', strains.flashlight);
-                    updateWithOffset('speed', strains.speed);
+                    updateWithOffset('aim', this.diffStrains.aim);
+                    // TODO:: remove aimNoSliders
+                    updateWithOffset('aimNoSliders', new Float64Array());
```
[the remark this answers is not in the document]
storycraft: no lol

packages/tosu/package.json
```diff
@@ -13,9 +13,9 @@
         "compile:linux": "pnpm run genver && pnpm run ts:compile && pnpm run compile:prepare-htmls && pkg --output dist/tosu --debug --config pkg.linux.json --compress brotli dist/index.js"
     },
     "dependencies": {
-        "@kotrikd/rosu-pp": "3.1.0",
         "@tosu/common": "workspace:*",
         "@tosu/ingame-overlay-updater": "workspace:*",
+        "@tosu/pp": "github:tosuapp/lazer-calculator#main&path:/lib",
```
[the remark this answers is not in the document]
storycraft: Pinned by lockfile

packages/tosu/src/memory/stable.ts
```diff
@@ -1285,18 +1280,15 @@ export class StableMemory extends AbstractMemory<OsuPatternData> {
...
-            good: 0,
-            ok: this.process.readShort(entry + 0x88),
-            meh: this.process.readShort(entry + 0x8c),
-            miss: this.process.readShort(entry + 0x92),
-
-            sliderTailHit: 0,
-            smallTickHit: 0,
-            largeTickHit: 0
-        };
+        const statistics = fromLegacyHitResults(
```
KotRikD: I don't like how it's done, maybe use named parameters function instead of args based?

KotRikD:
```js
const someFunc = ({ a, b }: { a: number; b: number; } ) => {}
```

packages/tosu/src/states/beatmap.ts
```diff
@@ -600,34 +606,29 @@ export class BeatmapPP extends AbstractState {
...
 
                 case 3:
-                    strainsAmount = strains.strains?.length || 0;
+                    strainsAmount = this.diffStrains.strains.length || 0;
                     break;
             }
 
-            const sectionOffsetTime = strains.sectionLength;
+            // Current lazer impl hardcodes each sections to 400
+            const sectionOffsetTime = 400;
```
KotRikD: maybe up it as a const and some comment where it's in osu code hardcoded? Because we apperantely will lose knowledge about this after sometime

packages/tosu/src/states/resultScreen.ts
```diff
@@ -126,84 +126,79 @@ export class ResultScreen extends AbstractState {
...
-                fcCalcOptions.n100 = 0;
-                fcCalcOptions.n50 = 0;
+                fcCalcOptions.greats = this.statistics.great;
+                fcCalcOptions.goods = this.statistics.good;
+                fcCalcOptions.oks = 0;
+                fcCalcOptions.mehs = 0;
                 fcCalcOptions.misses = 0;
-                delete fcCalcOptions.sliderEndHits;
-                delete fcCalcOptions.smallTickHits;
-                delete fcCalcOptions.largeTickHits;
```
KotRikD, line 190: bless god, we will not more have those constructions

packages/tosu/src/utils/hitResult.ts
```diff
@@ -0,0 +1,182 @@
...
+        100: getN100(mode, statistics),
+        300: statistics.great,
+        geki: getGeki(mode, statistics),
+        katu: getKatu(mode, statistics),
+        sliderEndHits: statistics.sliderTailHit,
+        smallTickHits: statistics.smallTickHit,
+        largeTickHits: statistics.largeTickHit
+    };
+
+    return hits;
```
KotRikD: `return {} as Hits3` , I guess no need to make another variable for this

packages/tosu/package.json
KotRikD: You want to hold it fresh on each build? Or what point of this lock?

packages/tsprocess/package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "node-addon-api": "^8.5.0",
     "@tosu/common": "workspace:*",
-    "node-gyp": "^11.5.0"
+    "node-gyp": "^12.2.0"
   },
```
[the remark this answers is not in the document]
KotRikD: we're okay with that

packages/tosu/package.json
storycraft: Not published to npm yet. Used sub directory because the repo is monorepo. Fixed to main branch to make sure it uses live pp calc(commit hash is pinned by pnpm-lock)

packages/tosu/src/states/beatmap.ts
storycraft: I think I should just move this to use lazer constant. So we don't need to update them

packages/tosu/src/utils/hitResult.ts
storycraft: Done

packages/tosu/src/memory/stable.ts
storycraft: Done

packages/tsprocess/package.json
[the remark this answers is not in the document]
storycraft: This is done because vs2026 won't work with node-gyp v11

packages/tosu/src/api/utils/buildResult.ts
```diff
@@ -172,12 +174,7 @@ export const buildResult = (instanceManager: InstanceManager): ApiAnswer => {
                 smooth: gameplay.playerHPSmooth
             },
             hits: {
-                300: gameplay.statistics.great,
-                geki: gameplay.statistics.perfect,
-                100: gameplay.statistics.ok,
-                katu: gameplay.statistics.good,
-                50: gameplay.statistics.meh,
-                0: gameplay.statistics.miss,
+                ...hits,
```
GabuTheDev, line 177: any reasons why we shallow copy here intead of directly passing `hits` ?

packages/tosu/src/api/utils/buildResultV2.ts
```diff
@@ -601,18 +602,10 @@ const buildLazerTourneyData = (
...
-                            50: client.score!.statistics.meh,
-                            0: client.score!.statistics.miss,
-                            sliderEndHits:
-                                client.score!.statistics.sliderTailHit || 0,
-                            smallTickHits:
-                                client.score!.statistics.smallTickHit || 0,
-                            largeTickHits:
-                                client.score!.statistics.largeTickHit || 0,
+                            ...toLegacyHits(
+                                client.score!.mode,
```
GabuTheDev, line 607: I know the previous code did the same, but I'd like to get rid of the non-null assertion operator.
Like, can we really ensure these properties exist? if so, why are they nullable to begin with?
if not, we should either fallback to a value or check beforehand.

packages/tosu/src/states/gameplay.ts
```diff
@@ -488,153 +494,158 @@ export class Gameplay extends AbstractState {
...
-
-                const difficulty = new rosu.Difficulty(commonParams);
-
+            if (isUpdate || !this.gradualPerformance) {
                 this.gradualPerformance =
-                    difficulty.gradualPerformance(currentBeatmap);
-                this.performanceAttributes = new rosu.Performance(
-                    commonParams
-                ).calculate(currentBeatmap);
+                    currentBeatmap.createGradualDifficultyCalculator();
```
GabuTheDev: Holy 33 char long method. Could we really not find another name for this?

packages/tosu/src/states/gameplay.ts
```diff
@@ -488,153 +494,158 @@ export class Gameplay extends AbstractState {
...
                 this.previousState = currentState;
             }
 
-            if (!this.gradualPerformance || !this.performanceAttributes) {
+            if (!this.gradualPerformance) {
                 wLogger.debug(
                     `%${ClientType[this.game.client]}%`,
                     `PP calc prerequisites missing:`,
-                    `gradual: ${this.gradualPerformance === undefined} - attributes: ${this.performanceAttributes === undefined}`
+                    `gradual: ${this.gradualPerformance === undefined}`
```
GabuTheDev: wouldn't it make more sense to log that the gradual performance is false? when it does not exist?

marking it as `gradual: true` when the message is `prerequisites missing` is uncommon.

packages/tosu/src/states/gameplay.ts
```diff
@@ -488,153 +494,158 @@ export class Gameplay extends AbstractState {
...
             );
+            // Check if hit object is actually ended
+            if (
+                passedObjects > 0 &&
+                beatmapPP.lazerBeatmap?.hitObjects[passedObjects - 1]
+            ) {
+                const hitObject =
+                    beatmapPP.lazerBeatmap.hitObjects[passedObjects - 1];
+                if ('endTime' in hitObject) {
+                    const endTime = (hitObject as unknown as IHasDuration)
```
GabuTheDev: fragile double type assertion... ugh

packages/tosu/src/states/resultScreen.ts
```diff
@@ -126,84 +126,72 @@ export class ResultScreen extends AbstractState {
...
+                ignoreHits: this.statistics.ignoreHit || 0,
+                ignoreMisses: this.statistics.ignoreMiss || 0,
+                largeBonuses: this.statistics.largeBonus || 0,
+                smallBonuses: this.statistics.smallBonus || 0,
+                perfects: this.statistics.perfect,
+                greats: this.statistics.great,
+                goods: this.statistics.good,
+                oks: this.statistics.ok,
+                mehs: this.statistics.meh,
+                misses: this.statistics.miss
```
GabuTheDev, line 157: Not against the property namings here but they do sound weird. 
"perfects" and "oks" especially.

GabuTheDev (changes requested):
Nice PR overall.

packages/tosu/src/api/utils/buildResult.ts
storycraft: Because it needs additional properties.

packages/tosu/src/states/gameplay.ts
storycraft: `true` because gradual *is* missing. Idk I just followed previous convention.

packages/tosu/src/states/gameplay.ts
storycraft: typescript interface is not real. It's best thing I could do. This is should be safe because it follows lazer's interface name and javascript doesn't have overloading.

packages/tosu/src/states/gameplay.ts
storycraft: Renamed to `createGradualDifficulty`

packages/tosu/src/states/gameplay.ts
GabuTheDev: Since `if` now has a single condition, might as well directly mention that `The gradual performance calculator is missing!` (or something like that)

packages/tosu/src/states/gameplay.ts
GabuTheDev: True. Interfaces are erased at runtime, but typescript was made for static analysis and runtime safety.

The double assertion `as unknown as IHasDuration` is unnecessary and masks potential type mismatches. Since we are already using the `in` operator to verify the property at runtime, can't we can rely on TypeScript's structural typing? or a direct assertion?

```diff
- if ('endTime' in hitObject) {
-   const endTime = (hitObject as unknown as IHasDuration).endTime;
+ if (hitObject && 'endTime' in hitObject) {
+ const { endTime } = hitObject as IHasDuration;
```

packages/tosu/src/states/beatmap.ts
storycraft: Moved to tosu/pp. Correct impl would be using `StrainSkill::SectionLength` but it's inaccessible. Not sure using patch would be good idea here.

Currently still hardcoded. But section offset time for some strains using non default values are correct now.

packages/server/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "rosu-pp-js": "4.0.1",
+    "@tosu/pp": "github:tosuapp/lazer-calculator#main&path:/lib",
```
KotRikD: Swap to release version?

packages/tosu/src/states/gameplay.ts
```diff
@@ -488,153 +494,157 @@ export class Gameplay extends AbstractState {
...
                     `%${ClientType[this.game.client]}%`,
                     `PP calc prerequisites missing:`,
-                    `gradual: ${this.gradualPerformance === undefined} - attributes: ${this.performanceAttributes === undefined}`
+                    `gradual: ${this.gradualPerformance === undefined}`
                 );
                 return;
             }
 
-            const passedObjects = calculatePassedObjects(
+            let passedObjects = calculatePassedObjects(
```
KotRikD: const?

KotRikD: I don't see any value changes

KotRikD (changes requested):

packages/tosu/src/states/gameplay.ts
storycraft: It was used on line 529 but I moved entire logic into `calculatePassedObjects` anyway

packages/tosu/src/states/gameplay.ts
storycraft: Just noticed the entire if body is unreachable due to previous if statement. It can be safely removed.

packages/tosu/src/api/utils/buildResultV2.ts
storycraft: This is better to be handled on other pr with separated issue for dx.

packages/tosu/src/states/resultScreen.ts
storycraft: I have no idea for better naming. I will consider changing if you give suggestions.

packages/server/package.json
storycraft: done

KotRikD (approved):

packages/tosu/src/utils/hitResult.ts
```diff
@@ -0,0 +1,172 @@
+import type { Hits2, Hits3 } from '@/api/types/v2';
```
GabuTheDev, line 1: Could we not have this whole `Hits2` `Hits3` `Hits99`?

We can use typescript's `Pick` or `Omit` to derive from the original.

GabuTheDev (approved):
Besides my comment regarding imports from `api/types/v2`, code-wise, it seems alright.

PR #600, opened 2026-05-03
feat: implemenet reading matchmaking stats

KotRikD:


packages/tosu/src/api/types/v2.ts
```diff
@@ -190,6 +190,14 @@ export interface NumberName {
     name: string;
 }
 
+export interface MatchmakingStats {
+    rating: number;
+    rank: number | null;
+    plays: number;
+    wins: number;
+    isProvisional: boolean;
+}
```
GabuTheDev, line 199: They introduced "seasons" with this new Ranked Play. (currently "season 0")
Is there any way to fetch that from the client?
as in, have the name of the season provided here?

KotRikD: Yeah, we can read it, but it seems doesn't have like history, only current one, soooo probably not interesting value

PR #601, opened 2026-05-03
feat: rosu pp update

KotRikD:


packages/tosu/src/states/beatmap.ts
```diff
@@ -427,7 +427,7 @@ export class BeatmapPP extends AbstractState {
                         mods: sanitizeMods(currentMods.array),
                         accuracy: acc,
                         lazer: this.game.client === ClientType.lazer,
-                        hitresultPriority: HitResultPriority.Fastest
```
GabuTheDev, line 430: Why are we changing this all of a sudden?
We were using Fastest preset because of a bug. (can't remember which. iirc it was something with delayed data)

PR #608, opened 2026-05-26
fix: update vtable on 2026.525.0

KotRikD:


packages/tosu/src/memory/lazer.ts
cyperdark, line 513: what about making this one as `number[]` to check for all previous and current vtables?

to support backwards compatibility for a 2-3 stable lazer releases?

cyperdark (changes requested):

packages/tosu/src/memory/lazer.ts
KotRikD: no.

KotRikD: (i will probably put vtable into offsets.json)

PR #609, opened 2026-05-30
fix: incorrect pp value when lazer CL mod or client accuracy error

storycraft:
Fix incorrect pp value when on lazer CL or when client provides wrong accuracy value.

* Added manual accuracy calculation for some weird case
* https://github.com/tosuapp/lazer-calculator/pull/1


KotRikD:
lgtm, waiting for ck reply

cyperdark (approved):

PR #613, opened 2026-06-01
feat: Add Ranked Play data to WebSocket.

Urantij:
adds ranked play field for v2 api websocket response


packages/tosu/src/memory/stable.ts
```diff
@@ -1442,4 +1443,8 @@ export class StableMemory extends AbstractMemory<OsuPatternData> {
             return error as Error;
         }
     }
+
+    rankedPlay(): ILazerRankedPlay | 'not-ready' {
+        throw new Error('в стабиле нет ранкед плея..');
+    }
```
GabuTheDev: Would be ideal if we do not have cyrillic in errors.

packages/tosu/src/memory/lazer.ts
```diff
@@ -3669,6 +3695,141 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
...
+                this.offsets['osu.Game.Online.Multiplayer.MultiplayerRoom'][
+                    '<MatchState>k__BackingField'
+                ]
+        );
+
+        if (!state) {
+            return 'not-ready';
+        }
+
+        const maybeStage = this.process.readInt(
```
GabuTheDev: these `maybeX`, `maybeY` namings need to be dropped.
I understand you're not sure but we're not going to implement something that we just "hope it works".

packages/tosu/src/memory/lazer.ts
```diff
@@ -3669,6 +3695,141 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
         return { chat: chatItems, spectatingClients };
     }
 
+    rankedPlay(): ILazerRankedPlay | 'not-ready' {
+        // idk about this
```
GabuTheDev: let's not leave comments like this shall we? )))

packages/common/enums/osu.ts
```diff
@@ -410,3 +410,55 @@ export enum ManiaScrollingDirection {
     up = 0,
     down = 1
 }
+
```
GabuTheDev, line 413: Should be stated that these namings are directly extracted from the game's codebase.

packages/tosu/src/states/rankedplay.ts
```diff
@@ -0,0 +1,38 @@
...
+    @measureTime
+    updateState() {
+        try {
+            const play = this.game.memory.rankedPlay();
+            if (play === 'not-ready') {
+                this.infoExist = false;
+            } else {
+                this.infoExist = true;
+                this.info = play;
+            }
```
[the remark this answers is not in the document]
Urantij: i wanted to have latest rankedplay data even if does not exist anymore, but i guess such way of caching is unnecessary. if i should to clear data, when ranked play isnt present in memory anymore, infoExist field should be removed

packages/tosu/src/memory/stable.ts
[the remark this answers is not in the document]
Urantij: я забыу про это проста

packages/tosu/src/utils/tprocessExtensions.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function readNullableInt(
+    process: Process,
+    address: number
+): number | undefined {
+    // if (process.readByte(address + 4) === 0) {
+    //     return undefined;
+    // }
+    // return process.readInt(address);
+    if (process.readByte(address) === 0) {
```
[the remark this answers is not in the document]
Urantij: че он ваще сделал

packages/tosu/src/memory/types.ts
```diff
@@ -209,4 +211,21 @@ export type ILazerSpectator =
...
+export type ILazerRankedPlay = {
+    stage: LazerRankedPlayStage;
+    currentRound: number;
+    damageMultiplier: number;
+    starRating: number;
+    users: { id: number; info: ILazerRankedPlayUserInfo }[];
+    activeUserId: number | undefined;
+    winningUserId: number | undefined;
+};
+
```
GabuTheDev: I'd drop `Lazer` from the name as Ranked Play is only ever going to be on lazer.
`ILazerRankedPlayUserInfo` -> `IRankedPlayUser`
`ILazerRankedPlay` -> `IRankedPlayRoom` because that's what it is, a multiplayer room.

this way, `IRankedPlayRoom#starRating` makes more sense than `ILazerRankedPlay#starRating`.

and I see we have `ILazerRankedPlay#winningUserId` as "the winner when the match ends", meaning this will only get populated when the match is in the `Results` and/or `Ended` phase?

and I'd split `ILazerRankedPlay#users` into `ILazerRankedPlay#player` and `ILazerRankedPlay#opponent` as it's ever going to be two players, and besides, what id is to which? just a simplifying thought.

this way we're going to have `rankedPlay.opponent.life` vs `rankedPlay.users[1].life`.

packages/tosu/src/utils/tprocessExtensions.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function readNullableInt(
+    process: Process,
+    address: number
+): number | undefined {
+    // if (process.readByte(address + 4) === 0) {
+    //     return undefined;
+    // }
+    // return process.readInt(address);
```
GabuTheDev: any reason why this is commented out?

GabuTheDev (changes requested):
Besides memory part, where I honestly cannot be bothered to check, code-wise seems good.
There are some structure changes that I think would be needed but overall 👍🏼 .

GabuTheDev:
as a side note, we might want to have the player's cards read too, but if it's too much work, it's alright.

packages/tosu/src/states/rankedplay.ts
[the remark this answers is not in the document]
GabuTheDev: > i wanted to have latest rankedplay data even if does not exist anymore

ideally, the websocket should not emit data that does not exist anymore.
caching can be done by overlay developers if needed.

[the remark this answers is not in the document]
Urantij: ye, now you mention it, and i think it makes sense... how do i change the code in this state? xd

packages/tosu/src/memory/lazer.ts
Urantij: They are called "maybe", because in this method its both pulling data and validating it. Its maybe a stage value, if all the variables are valid. I just didnt come up with a better name, and i also feel i should not name them just "stage", because its not known if its a stage or not. But if you think its ok, then its ok

packages/tosu/src/memory/lazer.ts
Urantij: forgot to remove it, my bad

packages/common/enums/osu.ts
Urantij: Like some formally structured message about licensing? Idk how to make this, can i just rename them and remove comments? xd

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: i was testing it, because i was not sure about int? memory layout. found a working way and forgot to remove this

packages/tosu/src/memory/lazer.ts
GabuTheDev: Is the validation you do even good to begin with?
how do we know a ranked play match cannot reach 50 rounds?
or a better question is: why do we validate them? is there a possibility of them being wrong?

packages/common/enums/osu.ts
GabuTheDev: that's not what I meant :)

I wanted a simple comment saying something like: 
```
Extracted as-is from "https://github.com/ppy/osu/blob/master/osu.Game/Online/Multiplayer/MatchTypes/RankedPlay/RankedPlayStage.cs"
```

packages/tosu/src/states/rankedplay.ts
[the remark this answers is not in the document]
GabuTheDev: > how do i change the code in this state? xd

rankedPlay data should be available while in ranked play, once the player left the multiplayer room, this should be null / undefined.

packages/tosu/src/memory/types.ts
Urantij: I agree with `Lazer` removal, but i dont like `UserInfo` to `User` change, because these are two different entities. Same with `RankedPlayRoom`. There are `Room` entities, and this entity is about room, but its not a `Room`

I didnt exactly tested, when winning user id is populated, but i believe it happens after gameplay part end, after which one of the players die. Iam 90% sure its populated during `Ended` state, and maybe during last `Result` state.

Iam not 100% sure that your own entity is always at index 0 and opponent at index 1. I guess its possible to filter out your own and then grab the other one as opponent, but i will have to lookup user's id for this. Also i heard that 2v2 mode is something ranked play's team is considering, so i like one array for this

Urantij:
> as a side note, we might want to have the player's cards read too, but if it's too much work, it's alright.

its possible, and shouldnt be that hard, but i did a minimum amount of work to make my next project possible, so maybe someone else will do the rest

packages/tosu/src/states/rankedplay.ts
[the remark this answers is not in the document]
Urantij: > > how do i change the code in this state? xd
> 
> rankedPlay data should be available while in ranked play, once the player left the multiplayer room, this should be null / undefined.

I agree with you. What i was trying to say is.. how do i change the code, so its added to this pr.. like, how do i commit this fix? xd

packages/tosu/src/memory/types.ts
GabuTheDev: I looked over the lazer codebase and they call it a RankedPlayRoom, so I would like to go with that.
And regarding the users array, I guess you could be right to keep it that way.

packages/tosu/src/memory/lazer.ts
Urantij: > how do we know a ranked play match cannot reach 50 rounds?

In current ranked play state its not possible, because after each game someone is dealing at least 50k flat damage. So its 40 rounds, if both players lose one after another for 1 score diff

> why do we validate them? is there a possibility of them being wrong?

yes. that `state` field may have different classes depending on game's state. Possible classes are SinglePlayerState, MultiplayerState, someotherstateidontrememberState and RankedPlayState. To be sure, that we have RankedPlayState, we need to verify fields, that only present in RankedPlayState. If not some other state, this memory will have non sense values

packages/common/enums/osu.ts
Urantij: Sure

packages/tosu/src/memory/lazer.ts
GabuTheDev: if that's the case, can't we check the main source? as instanceof something? instead of individually checking each field.

because there could be an edge case where all of these return true, but they are not ranked play stats.
just asking.

packages/tosu/src/memory/types.ts
Urantij: > I looked over the lazer codebase and they call it a RankedPlayRoom, so I would like to go with that. And regarding the users array, I guess you could be right to keep it that way.

https://github.com/ppy/osu/blob/master/osu.Game/Online/Multiplayer/MatchTypes/RankedPlay/RankedPlayRoomState.cs
https://github.com/ppy/osu/blob/master/osu.Game/Online/Multiplayer/MultiplayerRoom.cs

`RankedPlayRoomState` is a class about this room's state, but not the room itself. You can compare it to `MultiplayerRoom` class, which have things like `RoomID` and `MultiplayerRoomState` (room state, not ranked play state, because ranked play states are different)

https://github.com/ppy/osu/blob/master/osu.Game/Online/Multiplayer/MultiplayerClient.cs
It has `Room` field on line 155, which is a `Room` entity, which hold `State` entity on line 51. So these are 2 different things.
For this `State` model to become a "room", it need to at least have roomid, because its really weird to have a room object without its own room's id

Urantij: So if you want it to become a `Room`, it should be more like
```
{
  Id: number;
  State: State;
}
```

packages/tosu/src/memory/lazer.ts
Urantij: Well... its reading a raw memory. if you know how to validate by object's header, which class is it, it would be cool. But the codebase, which does the same thing, implies that its not possible.

There are chances, that that room's state field holds some other state, and random, nonsense memory we are reading is holding exactly needed values to make it look like a valid ranked play state, but i think the chances are astronomically low. And its not like we have a choice, because RankedPlayState does not contain something like references to other objects, which can be easily validated. So we just read these values, and if they make sense, its 99.9% is ranked play

Urantij: Removed "maybe" part. I think its possible to understand this situation without such prefix

longnguyen2004:
Tiny nit: Returning a whole `"not-ready"` string when the data is unavailable is overkill, when you can just return `undefined` instead. Then you can get rid of https://github.com/Urantij/tosu/blob/72b2e85816316d0ab57d18938523611948ddcd03/packages/tosu/src/states/rankedplay.ts#L13-L17 and just assign it directly to the data payload.

packages/tosu/src/memory/lazer.ts
longnguyen2004: > if that's the case, can't we check the main source? as instanceof something? instead of individually checking each field.

You can't reimplement .NET type checking with just memory reading unfortunately

Urantij:
> Tiny nit: Returning a whole `"not-ready"` string when the data is unavailable is overkill, when you can just return `undefined` instead. Then you can get rid of https://github.com/Urantij/tosu/blob/72b2e85816316d0ab57d18938523611948ddcd03/packages/tosu/src/states/rankedplay.ts#L13-L17 and just assign it directly to the data payload.

yes, but i just copied this behavior from somewhere in tosu codebase

GabuTheDev:
Request a review when ready @Urantij

Urantij:
> Request a review when ready @Urantij

uhh, i dont know. I did all the settled changes, but we still have `RankedPlay` naming discussion open. If you insist on a different name, i'd like to change `RankedPlay` to `RankedPlayState`. Other than that, everything should be ready, i believe

GabuTheDev:
I'd still be looking into:
- https://github.com/tosuapp/tosu/pull/613/changes#r3334889519
- https://github.com/tosuapp/tosu/pull/613/changes#r3334889485
- https://github.com/tosuapp/tosu/pull/613/changes#r3334889457

Urantij:
> I'd still be looking into:
> 
> * https://github.com/tosuapp/tosu/pull/613/changes#r3334889519
> * https://github.com/tosuapp/tosu/pull/613/changes#r3334889485
> * https://github.com/tosuapp/tosu/pull/613/changes#r3334889457

iam genuinely lost. i just dont understand, what iam looking at, when i open these links.

https://github.com/tosuapp/tosu/pull/613/changes#r3334889519
infoExist does not exist in the code anymore. we discussed that tosu shouldnt cache rankedplay data, and its always returns current memory state. i dont understand what i need to look at

https://github.com/tosuapp/tosu/pull/613/changes#r3334889485
you dont like `// Extracted as-is from "https://github.com/ppy/osu/blob/101cfffb80577e1910cb741b496743c2427617fa/osu.Game/Online/Multiplayer/MatchTypes/RankedPlay/RankedPlayStage.cs"` comment? What comment i need to put there then?

https://github.com/tosuapp/tosu/pull/613/changes#r3334889457
this link is also opening discussion about required comment. i dont understand.

GabuTheDev:
oh great. they just do not redirect to where I wanted.
look at copilot's `high` priority comments.

packages/tosu/src/memory/stable.ts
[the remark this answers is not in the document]
Urantij: it does not make sense

packages/tosu/src/utils/tprocessExtensions.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function readSharpDictionaryIntToRef(
+    process: Process,
+    address: number
+): DictionaryIntToRefEntry[] {
+    const result = [];
+
+    const entriesArray = process.readIntPtr(address + 0x8 + 0x8);
+
+    const entriesLength = process.readInt(entriesArray + 0x8);
```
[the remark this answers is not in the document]
Urantij: In this codebase `address` is always not checked during first memory read. Its always is done either before or not at all

packages/tosu/src/memory/lazer.ts
```diff
@@ -3669,6 +3695,141 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
...
+            state +
+                this.offsets[
+                    'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayRoomState'
+                ]['<Users>k__BackingField']
+        );
+
+        const users = readSharpDictionaryIntToRef(
+            this.process,
+            usersDictionary
+        );
```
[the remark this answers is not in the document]
Urantij: it does not really matter at that point, because if its a real rankedplay state, everything will be fine, but if its a "fake" rankedplay state, everything will break anyway. but we can consider this as another validation step, which is kinda pointless, since its really unlikely to have 8 zeroes at random memory chunk. fixed with #613

packages/tosu/src/api/utils/buildResultV2.ts
```diff
@@ -142,6 +144,7 @@ export const buildResult = (instanceManager: InstanceManager): ApiAnswer => {
             playTime: global.gameTime,
             playCount: 0 // need counting
         },
+        rankedPlay: rankedPlay.info,
         settings: {
```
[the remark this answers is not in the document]
Urantij: fake and gay

Urantij:
everything is resolved, good night

packages/tosu/src/memory/index.ts
```diff
@@ -58,6 +59,7 @@ export abstract class AbstractMemory<M extends Record<string, number>> {
     abstract tourneyChat(messages: ITourneyManagerChatItem[]): ITourneyChat;
     abstract tourneyUser(): ITourneyUser;
     abstract leaderboard(mode: number): ILeaderboard;
+    abstract rankedPlay(): IRankedPlay | 'not-ready';
```
KotRikD: only `IRankedPlay` should be here

packages/tosu/src/memory/lazer.ts
```diff
@@ -162,7 +168,9 @@ export interface Offsets {
         '<client>k__BackingField': number;
     };
     'osu.Game.Online.Multiplayer.MultiplayerRoom': {
+        RoomID: number;
```
KotRikD, line 164: cover this in `''`

packages/tosu/src/memory/lazer.ts
```diff
@@ -387,6 +395,24 @@ export interface Offsets {
...
+        '<ActiveUserId>k__BackingField': number;
+        '<StarRating>k__BackingField': number;
+        '<WinningUserId>k__BackingField': number;
+    };
+    'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayUserInfo': {
+        '<Rating>k__BackingField': number;
+        '<Life>k__BackingField': number;
+        '<Hand>k__BackingField': number;
+        '<RatingAfter>k__BackingField': number;
+        DamageInfo: number;
```
KotRikD, line 393: same

packages/tosu/src/memory/types.ts
```diff
@@ -209,4 +211,21 @@ export type ILazerSpectator =
...
     | undefined;
 
+export type IRankedPlayUserInfo = {
+    rating: number;
+    life: number;
+    roundsWon: number;
+    damageMultiplier: number;
+};
+
+export type IRankedPlay = {
```
KotRikD:
```ts
export type IRankedPlay = {} | string | Error;
```
(maybe no need for error, but string for `not-ready` 100%)

packages/tosu/src/utils/tprocessExtensions.ts
KotRikD, line 1: Should be part of `tsprocess` package, not tosu

packages/tosu/src/utils/tprocessExtensions.ts
```diff
@@ -0,0 +1,44 @@
...
+    for (let i = 0; i < entriesLength; i++) {
+        const entrySize = 24;
```
KotRikD: Why it's constant?

packages/tosu/src/utils/tprocessExtensions.ts
```diff
@@ -0,0 +1,44 @@
...
+        const entry = entriesArray + 0x8 + 0x8 + i * entrySize;
```
KotRikD: I'm not sure, but isn't `0x8 + 0x10` ?

KotRikD (changes requested):

packages/tosu/src/memory/index.ts
Urantij: should rankedPlay() throw error if its not ready?

Urantij: oh the type needs to include string

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: you mean `entrySize`? because this struct's byte size is 24.

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: The code is reading users dictionary correctly, but i double checked, and dictionary's memory layout is 8 bytes header, 4 bytes length, 4 bytes padding and then entry elements (each 24 bytes)

Urantij: i meant entries array memory layout

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: its kinda not intuitive for me how to do it. i did this as an extension, because the code is linking process from /tsprocess/lib/process and if i modify /tsprocess/src/process, it does not link properly. should i still add these methods to process.ts and recompile?

packages/tosu/src/memory/lazer.ts
Urantij: uhhh, `pnpm run prettier:fix` removes `'`. there are also a lot of variables that dont include `'`, so iam not sure if this is really needed

packages/tosu/src/memory/lazer.ts
Urantij: same problem with prettier

packages/tosu/src/memory/types.ts
Urantij: added | string

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: did exactly that. if it is what needed, then its done

packages/tosu/src/utils/tprocessExtensions.ts
Urantij: i just realized that you asked why i made this external constant, and not why i made a constant value constant, lol.
While memory walking is kinda self explaining, i felt like it should be stated why we are multiplying index by 24

Urantij: i get it, it should be provided as argument :D

Urantij: oh wait no it shouldnt

packages/tosu/src/memory/lazer.ts
KotRikD: lmao

KotRikD:
I will look into this pr on next week to merge (atleast need verify that our dumper will satisfy those fields)

packages/tosu/src/memory/lazer.ts
```diff
@@ -3669,6 +3691,138 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
...
+                state +
+                    this.offsets[
+                        'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayRoomState'
+                    ]['<ActiveUserId>k__BackingField']
+            ),
+            winningUserId: this.process.readNullableInt(
+                state +
+                    this.offsets[
+                        'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayRoomState'
+                    ]['<WinningUserId>k__BackingField']
```
cyperdark: i would prefer if `activeUserId` and `winningUserId` was moved to `user.isActive` and `user.isWinner`

packages/tosu/src/states/rankedplay.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    @measureTime
+    updateState() {
+        try {
+            const play = this.game.memory.rankedPlay();
+            if (play === 'not-ready') {
+                this.info = undefined;
+            } else {
+                this.info = play;
+            }
```
cyperdark:
```suggestion
           this.info = play !== 'not-ready' ? play : undefined;
```

packages/tosu/src/memory/lazer.ts
```diff
@@ -3669,6 +3691,138 @@ export class LazerMemory extends AbstractMemory<LazerPatternData> {
...
+                                'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayUserInfo'
+                            ]['<RoundsWon>k__BackingField']
+                    ),
+                    damageMultiplier: this.process.readDouble(
+                        u.address +
+                            this.offsets[
+                                'osu.Game.Online.Multiplayer.MatchTypes.RankedPlay.RankedPlayUserInfo'
+                            ]['<DamageMultiplier>k__BackingField']
+                    )
+                }
```
cyperdark: no need for `info` object, you can put them directly in user:
user.id
user.rating
user.life
etc.

cyperdark (changes requested):

packages/tosu/src/memory/lazer.ts
Urantij: makes sense, i just followed the game memory layout, but it can be simplified

packages/tosu/src/memory/lazer.ts
Urantij: i dont like this. it adds unnecessary data, which is hard for me to see as usable. on client side, if you want to filter winner user, you will use `users.filter(u => u.id === winningUserId)` or `users.filter(u => u.isWinner)` which basically the same work. if you are already storing current users, and want to check, if the user is winning, you will have to filter out the user you need from the server, and then check if they are winning, instead of just checking `winninguserid`
this change feels dirty to me

cyperdark: okay, then move them above everything

Urantij: it also feels weird, because everything else is always presented in the model, while `activeUserId` `winningUserId` can and will be undefined

cyperdark: make it: `value || 0`

Urantij: я прально понимаю идею, что ты хош насильно популировать поле, чтобы оно типа всегда было в моделе, чтобы его можно было закинуть выше? и усложнить логику, что ноль это на самом деле не айди, а его отсутствие

cyperdark: если `value || 0` то проверка будет только number == number, а в случае если этих полей нету тебе нужно будет делать проверку на их существование. Да и насколько я понял тот кто будет делать оверлей в любом случае будет чекать `if activeUserId != profile.user.id then dont_show_anything` или я что то упустил?

а проверка в массиве вообще не изменится `users.find(user => user.id == rankedPlay.activeUserId)`

Urantij: не совсем так. тип поля будет только `number`, то есть нужен как минимум коммент, что 0 это отсутствие значения, в то время как `number | undefined` сразу даёт понять логику.
если тебе нужно узнать, есть ли победитель, тебе нужно делать проверку `winningUserId !== 0`, либо прочёсывать каждый раз массив юзеров, хотя у тебя винюзерайди 0, что тож странно. это переусложнение, и я не понимаю, зач

Urantij: ну и оно будет добавлять по сути лишнее поле в сериализованную модель, что тож кабутабы не оч. если так надо, чтобы опциональное поле всегда было представлено, можно хотя бы нулом его сделать тада

Urantij: короче, если обязательно нужно, чтобы эти два поля были выше, лучше уж сделать их выше и `number | undefined`. среди уже существующих в коде типов встречается поля как `type | null` так и `type | undefined`, так что сам по се андфаейнд не должен быть проблемой

cyperdark: они всегда должны быть здесь, + в тосу либо 0 или -1 значит unknown state

например если не залогинен в осу profile.id == -1
resultsScreen.scoreId == 0 если скор оффлайн

так что я лучше бы использовал 0 или -1

Urantij: окей босс, мне это не нравится, но будь по твоему. выбирай, -1 или 0

cyperdark: -1

Urantij: что не пожелается всё сбудется

cyperdark (dismissed):

PR #627, opened 2026-06-17
fix: Fix graph offset on maps starting and ending with spinner

cyperdark:
iirc rosu was ignoring spinners in strains but in lazer it returns 0, which on maps that starts and ends with spinner causes a significant offsets

image
image

### DIFF
--- a/packages/tosu/src/states/beatmap.ts
+++ b/packages/tosu/src/states/beatmap.ts
@@ -11,7 +11,7 @@ import {
     type StrainsData
 } from '@tosuapp/lazer-calculator';
 import fs from 'fs';
-import { HitType, Beatmap as ParsedBeatmap, TimingPoint } from 'osu-classes';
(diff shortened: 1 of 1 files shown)


packages/tosu/src/states/beatmap.ts
```diff
@@ -502,11 +499,6 @@ export class BeatmapPP extends AbstractState {
                 this.timings.firstObj = Math.round(
                     this.lazerBeatmap.hitObjects.at(0)?.startTime ?? 0
```
storycraft, line 500:
```suggestion
                    this.lazerBeatmap.hitObjects[1]?.startTime ?? 0
```
First hit object always have 0 value.
https://github.com/ppy/osu-tools/blob/d82cf2a986218f80c18dea5cc1399133d4a8586d/PerformanceCalculatorGUI/Screens/SimulateScreen.cs#L740

KotRikD (approved):

PR #630, opened 2026-06-20
fix: Improve `failed to fetch offsets` message

cyperdark:


packages/tosu/src/instances/lazerInstance.ts
```diff
@@ -98,10 +98,17 @@ export class LazerInstance extends AbstractInstance {
                             payload: text,
                             defaultValue: null
                         });
-                        if (json === null) continue;
+                        if (json === null) {
+                            wLogger.debug(
+                                `Malformed offsets on %${host}%:`,
```
GabuTheDev: "Malformed" ????????? ))))

KotRikD (approved):

PR #635, opened 2026-06-23
feat: dynamic pp calculator switching

storycraft:
# Todo
* [x] Dashboard settings
* [x] Move calculator typings and imports to separated package.


packages/pp-module-loader/src/registry.ts
```diff
@@ -0,0 +1,40 @@
+import { NpmRegistry } from 'npm-registry-sdk';
```
cyperdark: i would prefer if we dont use this whole npm package just for one call, instead we can have one fetch request with types


```ts
import { wLogger } from './logger.js';

interface npmPackageInfo {
    _id: string;
    _rev: string;
    name: string;
    'dist-tags': {
        latest: string;
        'pp-dev': string;
    };
    versions: Record<
        string,
        {
            name: string;
            version: string;
            author: {
                name: string;
                email: string;
            };
            _id: string;
            maintainers: {
                name: string;
                email: string;
            }[];
            homepage: string;
            bugs: {
                url: string;
            };
            dist: {
                shasum: string;
                tarball: string;
                fileCount: number;
                integrity: string;
                signatures: {
                    sig: string;
                    keyid: string;
                }[];
                unpackedSize: number;
            };
            _from: string;
            _npmUser: {
                name: string;
                email: string;
            };
            _resolved: string;
            _integrity: string;
            repository: {
                url: string;
                type: string;
            };
            _npmVersion: string;
            description: string;
            directories: {};
            _nodeVersion: string;
            _hasShrinkwrap: boolean;
            _npmOperationalInternal: {
                tmp: string;
                host: string;
            };
        }
    >;
    time: Record<string | 'created' | 'modified', string>;
    bugs: {
        url: string;
    };
    author: {
        name: string;
        email: string;
    };
    license: string;
    homepage: string;
    repository: {
        url: string;
        type: string;
    };
    description: string;
    maintainers: {
        name: string;
        email: string;
    }[];
    readme: string;
    readmeFilename: string;
}

export interface packageInfo {
    digest: `${string}:${string}`;

    name: string;
    version: string;

    files: number;
    size: number;

    url: string;
}

export async function npmVersions(packageName: string): Promise<
    | {
          status: 'error';
          message: string;
      }
    | {
          status: 'success';
          message: packageInfo[];
      }
> {
    try {
        const request = await fetch(
            `https://registry.npmjs.org/${packageName}`
        );
        if (!request.ok)
            return {
                status: 'error',
                message: `${request.status} - ${request.statusText}`
            };

        const json = (await request.json()) as npmPackageInfo;

        return {
            status: 'success',
            message: Object.values(json.versions)
                .map((item) => {
                    const keyid =
                        item.dist.signatures.find((r) =>
                            r.keyid.startsWith('SHA256:')
                        )?.keyid || '';

                    if (keyid === '') return null;

                    const [algorithm, base64] = keyid.split(':');
                    return {
                        digest: `${algorithm}:${Buffer.from(base64).toString('hex')}` as const,

                        name: item.name,
                        version: item.version,

                        files: item.dist.fileCount,
                        size: item.dist.unpackedSize,

                        url: item.dist.tarball
                    };
                })
                .filter((item) => item != null)
        };
    } catch (error) {
        wLogger.debug('npmVersions', error);
        return {
            status: 'error',
            message: (error as Error).message
        };
    }
}
```


just and example:
```ts
const versions = await npmVersions('realm');
if (versions.status == 'error') {
	console.error('getting versions failed', versions);
	process.exit(1);
};


const selected = versions.message.find(r => r.version == version) || versions.message.at(-1);
if (typeof selected != 'object') {
	console.error('wrong version type', typeof selected, [selected]);
	process.exit(1);
};


const label = `${selected.name}@${selected.version} - ${selected.digest}`;
console.log('version found', label);



const file_path = path.join(cache_directory, path.basename(selected.url));
// if (fs.existsSync(realm_file_path)) await fs.promises.rm(realm_file_path).catch(null);
if (!fs.existsSync(file_path)) {
	console.log('downloading', label);

	const download = await download_file(selected.url, file_path);
	if (download.status == 'error') {
		console.error(`downloading "${selected.url}" failed`, download);
		process.exit(1);
	};
};


const verify = await verify_file(selected.digest, file_path);
if (verify.status == 'error') {
	console.error(`verify failed`, verify);
	process.exit(1);
};

console.log('verified', label, '\n');
```

packages/pp-module-loader/src/downloader.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (await fs.stat(folderPath).catch(() => false)) {
+        return folderPath;
+    }
+
+    await fs.mkdir(CALCULATOR_FOLDER, { recursive: true });
+    await downloadFile(
+        `https://registry.npmjs.org/${fullPackageName}/-/${packageName}-${version}.tgz`,
+        archivePath
+    );
+    try {
```
cyperdark: with module i sent above you can just pass url and digest to verify it's actually the file it downloaded

example: https://github.com/tosuapp/tosu/blob/master/packages/updater/index.ts#L154-L160

packages/pp-module-loader/src/downloader.ts
```diff
@@ -0,0 +1,48 @@
+import { downloadFile } from '@tosu/common';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import * as tar from 'tar';
+
+import { getFullPrebuiltPackageName, getPrebuiltPackageName } from './package';
+
+const CALCULATOR_FOLDER = 'pp-calculator';
```
cyperdark: tosu already has cache folder, so we can put only .node files in this path:
```ts
${getCachePath}/calculators or pp-calculators/${packageVersion}.node
```

packages/pp-module-loader/src/downloader.ts
```diff
@@ -0,0 +1,48 @@
...
+        return folderPath;
+    }
+
+    await fs.mkdir(CALCULATOR_FOLDER, { recursive: true });
+    await downloadFile(
+        `https://registry.npmjs.org/${fullPackageName}/-/${packageName}-${version}.tgz`,
+        archivePath
+    );
+    try {
+        await fs.mkdir(folderPath, { recursive: true });
```
cyperdark, line 39: will fail if folder already exists

packages/pp-module-loader/src/downloader.ts
```diff
@@ -0,0 +1,48 @@
...
+        await tar.x({
+            file: archivePath,
+            cwd: folderPath,
+            strip: 1
+        });
```
cyperdark, line 44: yesterday i had to unpack `.tar.gz` & `.tgz`, for that i used system tar with calling execFile, and as far as i know it's installed by default on windows and linux


```ts
import { execFile } from 'node:child_process';
import { promisify } from 'node:util';
import path from 'path';


async function extractTar(file_path: string, destination: string, args: string[]): Promise<string | Error> {
	try {
		const tgz_path = path.resolve(file_path);
		const dest_path = path.resolve(destination);

		const result = await promisify(execFile)('tar', ['-xzf', `${tgz_path}`, '-C', `${dest_path}`, ...args]);
		if (result.stderr) return new Error(result.stderr);

		return result.stdout;
	} catch (error: any) {
		return error.code === 'ENOENT'
			? new Error('The "tar" command is not found in your system PATH. Please ensure it is installed.')
			: error;
	};
};
```


example of usage:
```ts
const extract = await extractTar(file_path, cache_directory, ['--strip-components=2']);
if (extract.status == 'error') {
	console.error(`extracting failed`, extract);
	process.exit(1);
};
```

cyperdark (changes requested):

packages/pp-module-loader/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "name": "@tosu/pp-module-loader",
+    "version": "0.1.0",
+    "description": "Load and manage pp calculators",
+    "main": "src/index.ts",
+    "scripts": {},
+    "dependencies": {
+        "@tosu/common": "workspace:*",
+        "@tosuapp/lazer-calculator": "0.4.0-20260616.0",
+        "npm-registry-sdk": "^1.2.1",
+        "semver": "^7.7.3",
```
cyperdark: oops forgot about this one, pnpm has catalogs to reuse packages https://pnpm.io/catalogs

probably out of scope for this PR, up to you

packages/pp-module-loader/src/downloader.ts
storycraft: Only when `recursive: false`. No errors on `recursive: true` aka mkdirs.

packages/pp-module-loader/src/downloader.ts
storycraft: Addressed in 39aa24fc54fb0ecaee305a220089586ea022f05f

packages/pp-module-loader/src/downloader.ts
cyperdark: kk

packages/pp-module-loader/src/registry.ts
storycraft: Addressed in 9c846e3a8ccfb54b00811f0bceaeeb0c1a2dcc67

packages/pp-module-loader/src/downloader.ts
storycraft: No idea if it's good idea to rely on os command

packages/pp-module-loader/src/index.ts
```diff
@@ -0,0 +1,110 @@
...
+                throw new Error(msg);
+            }
+
+            const pkg = await onlinePpRegistry.fetch(module.version);
+            if (!pkg) {
+                wLogger.warn(
+                    '[calculator]',
+                    `Unable to fetch release information "${module.version}" from registry.`
+                );
+            }
```
cyperdark: code above throw new Error, but not in there

packages/pp-module-loader/src/downloader.ts
cyperdark: @KotRikD

packages/pp-module-loader/package.json
cyperdark: forgot to mention, this one was for semver

packages/pp-module-loader/src/index.ts
storycraft: release version should work on offline if calculator is already downloaded. For `dist-tag`, it's impossible to know which version is latest on offline

cyperdark: fuck i forgot to attach image 
image

storycraft: 1. `downloadCalculator` downloads pp calculator and returns its path. First it checks if folder exists, otherwise it downloads using dist information.
2. For `release`, it can still load offline version if it's already downloaded.
3. For `dist-tag` (eg. pp-dev or main), we must resolve it into specific version. If it fails, throw right away.

I will update the code so it's less confusing

packages/pp-module-loader/src/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    events: new EventEmitter<{ changed: [] }>(),
+
+    get current(): LazerCalculator {
+        return currentCalculator;
+    },
+
+    async load(module: PpModule): Promise<void> {
+        try {
+            currentCalculator = await resolveCalculator(module);
+            wLogger.info('[calculator]', 'Successfully loaded calculator.');
```
GabuTheDev: I don't want to bother, but I'd pretty much like if we could stick with the newer log messages.
as per https://github.com/tosuapp/tosu/commit/2048c4cc9cbe47f9b3ed3ecf36c52141c868bf64

packages/pp-module-loader/src/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    switch (module.type) {
+        case 'internal': {
+            wLogger.info('[calculator]', 'Using internal calculator');
+            return internalCalculator;
+        }
+
+        case 'dist-tag': {
+            wLogger.info(
+                '[calculator]',
+                `Using dist-tag "${module.tag}" calculator`
```
GabuTheDev: These log messages really do not tell anything for the end user.
Either make them end user friendly, or move them to debug.

packages/pp-module-loader/src/index.ts
storycraft: Yeah the message was from old lazer pp work from ck. Let me standardize them

PR #645, opened 2026-07-01
fix: fc pp skill value

storycraft:


storycraft:
ohsh (wrong base branch lol)

cyperdark (approved):

PR #650, opened 2026-07-05
feat: add room data to websocket

Urantij:
Will add room model, which exist in memory, when the player is in a ranked play match or just lounge lobby.

This allows to fetch data about users like username, id, country and avatarurl in a current multiplayer room


Urantij:
xd well idk how it was merged with non lint code but ээээто былааа не я не я не я

KotRikD:
@codex review

storycraft (changes requested):
Now we require all types to be type import when the values are not used. So please change them.

KotRikD:
example

```ts
import { type <TYPE> } from './<TYPE>'
```

Urantij:
iam not sure if this is the right thing i needed to do, since some of these imports were not imported as type, but vs code was complaining so idk

Urantij:
do i need to resolve this conflict or you just dont have time to move this pr?

KotRikD:
> do i need to resolve this conflict or you just dont have time to move this pr?

Don't have time rn, will look on this week

KotRikD (approved):

PR #651, opened 2026-07-06
chore: cleanup tsconfig and prepare for esm

storycraft:
* update `typescript` to v6
* update to `nodenext` module and module resolution
* replace `ts-node` to `tsx` to support esm
* remove remaining `tsc` process in `@tosu/updater`
* introduce common tsconfig: `tsconfig.base.json`
* cleanup tsconfig
* esmify `postBuild.ts` to `postBuild.mts`


packages/tosu/tsconfig.json
```diff
@@ -1,31 +1,25 @@
...
-        "@/*": ["src/*"]
-      },
-      "typeRoots": [
-        "./src/@types"
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "rootDir": "../..",
+    "allowJs": true,
+    "resolveJsonModule": true,
+    "emitDecoratorMetadata": true,
```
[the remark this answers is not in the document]
storycraft: it's deprecated on v6 and properly configured to relative path.

packages/tosu/src/postBuild.mts
GabuTheDev, line 1: Why is this a `.mts`? Is changing the project module target to `ES Module` directly breaking the whole thing?

storycraft: Yes it breaks whole things. Also without using mts, we cannot use nodenext module resolution due to resedit.

storycraft: ncc doesn't properly support esm. We would need to find another bundler for that

GabuTheDev (dismissed):

packages/tosu/src/postBuild.mts
cyperdark: maybe something like rolldown: https://rolldown.rs/guide/notable-features

storycraft:
There was remaining v5 and peer dependency issues so I added those fixes. Also I updated DEVELOPMENT.md

GabuTheDev (approved):

PR #655, opened 2026-07-09
feat: Move to Rolldown bundler + esmify codebase

storycraft:
1. Switched tsconfig module resolution to bundler.
2. Replaced ncc to rolldown
3. All workspace packages are now native esm(`"type": "module"`).
4. Effectively removing all differences between dev and prod environment. It runs on bundled file + source map to match with production env. Resolving pkgAssetsPath hack and allows to use bundler post processing stuffs.

* Blocked by #651. Do not merge it before that pr.

### DIFF
--- a/packages/common/utils/directories.ts
+++ b/packages/common/utils/directories.ts
@@ -74,7 +74,6 @@ export function getCachePath() {
 }
 
 export function getProgramPath() {
-    if ('pkg' in process) return path.dirname(process.execPath);
     return process.cwd();
 }
 
--- a/packages/common/utils/directories.ts
+++ b/packages/common/utils/directories.ts
@@ -74,6 +74,7 @@ export function getCachePath() {
 }
 
 export function getProgramPath() {
+    if ('pkg' in process) return path.dirname(process.execPath);
     return process.cwd();
 }
 
--- a/packages/server/router/assets.ts
+++ b/packages/server/router/assets.ts
@@ -4,10 +4,7 @@ import path from 'path';
 
 import { Server, getContentType } from '../index';
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
 
 export default function buildAssetsApi(server: Server) {
     server.app.route(/^\/assets\/(?<filePath>.*)/, 'GET', (req, res) => {
--- a/packages/server/router/index.ts
+++ b/packages/server/router/index.ts
@@ -1,5 +1,5 @@
 import {
-    ConfigBinding,
+    type ConfigBinding,
     ConfigManager,
     JsonSafeParse,
     downloadFile,
@@ -26,15 +26,12 @@ import {
     getLocalCounters,
     saveSettings
 } from '../utils/counters';
-import { ISettings } from '../utils/counters.types';
+import { type ISettings } from '../utils/counters.types';
 import { directoryWalker } from '../utils/directories';
 import { parseCounterSettings } from '../utils/parseSettings';
 import { generateReport, generateReportHTML } from '../utils/report';
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
 
 export default function buildBaseApi(server: Server) {
     server.app.route('/json', 'GET', (req, res) => {
--- a/packages/server/utils/counters.ts
+++ b/packages/server/utils/counters.ts
@@ -12,7 +12,7 @@ import path from 'path';
 import semver from 'semver';
 
 import { getContentType } from '../utils';
-import { ICounter, bodyPayload } from './counters.types';
+import type { ICounter, bodyPayload } from './counters.types';
 import {
     authorHTML,
     authorLinksHTML,
@@ -35,16 +35,7 @@ import {
 } from './htmls';
 import { parseCounterSettings } from './parseSettings';
 
-/**
- * ТАК КАК БЛЯТЬ У НАС В ЖЫЭСЕ
- * НЕ ПРИДУМАЛИ НОРМАЛЬНО ПАКЕТИРОВАТЬ ГОВНО БЕЗ ВЕБПАКА
- * ИДЕМ И ПИЛИМ КОСТЫЛИ
- * kys js!
- */
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
 
 function splitTextByIndex(text: string, letter: string) {
     const index = text.indexOf(letter);
--- a/packages/server/utils/report.ts
+++ b/packages/server/utils/report.ts
@@ -112,10 +112,7 @@ export async function generateReport(instanceManager: any): Promise<Report> {
     };
 }
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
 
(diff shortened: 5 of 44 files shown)


packages/tosu/rolldown.config.mjs
```diff
@@ -0,0 +1,59 @@
...
+            this.emitFile({
+                type: 'asset',
+                fileName: 'rosu_pp_js_bg.wasm',
+                source: await fs.readFile(
+                    path.join(rosuDir, 'rosu_pp_js_bg.wasm')
+                )
+            });
+        }
+    };
+}
```
GabuTheDev, line 59: Mind explaining a little here?

cyperdark: rolldown doesn't copy .wasm/.node, and you need to make plugin for workaround

storycraft: rolldown can only detect wasm loaded by import so this is fix for that.

packages/common/utils/directories.ts
```diff
@@ -74,7 +74,6 @@ export function getCachePath() {
 }
 
 export function getProgramPath() {
-    if ('pkg' in process) return path.dirname(process.execPath);
     return process.cwd();
```
GabuTheDev, line 78: [ERROR] Removing the `pkg` in process check from `getProgramPath` causes a regression in the packaged executable. Under `pkg` packaging, `process.cwd()` returns the folder from which the user ran the command. This will cause files like databases, configs (`.env`), cache, and update assets to be written to and read from arbitrary directories rather than next to the executable itself (`process.execPath`).

```suggestion
export function getProgramPath() {
    if ('pkg' in process) return path.dirname(process.execPath);
    return process.cwd();
}
```

packages/server/router/assets.ts
```diff
@@ -4,10 +4,7 @@ import path from 'path';
 
 import { Server, getContentType } from '../index';
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
```
GabuTheDev, line 7: [WARNING] Changing `pkgAssetsPath` to look for a local `assets` directory relative to `import.meta.dirname` works for the bundled output (as files are copied next to the final `dist/index.js`), but it breaks local unbundled execution (e.g. running via `ts:run`). When unbundled, `import.meta.dirname` resolves to `packages/server/router` or `packages/server/utils`, which prevents the server from locating the actual assets directory at `packages/server/assets`. Adding an existence fallback ensures that both unbundled dev execution and bundled prod environments resolve assets correctly.

```suggestion
const pkgAssetsPath = fs.existsSync(path.join(import.meta.dirname, 'assets'))
    ? path.join(import.meta.dirname, 'assets')
    : path.join(import.meta.dirname, '../assets');
```

packages/server/router/index.ts
```diff
@@ -26,15 +26,12 @@ import {
...
+import { type ISettings } from '../utils/counters.types';
 import { directoryWalker } from '../utils/directories';
 import { parseCounterSettings } from '../utils/parseSettings';
 import { generateReport, generateReportHTML } from '../utils/report';
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
```
GabuTheDev, line 40: [WARNING] Applying the same asset path fallback check to `packages/server/router/index.ts` to support both dev (unbundled) and production (bundled) environments.

```suggestion
const pkgAssetsPath = fs.existsSync(path.join(import.meta.dirname, 'assets'))
    ? path.join(import.meta.dirname, 'assets')
    : path.join(import.meta.dirname, '../assets');
```

GabuTheDev (commented):
## 📋 Review Summary

This pull request transitions the workspace's build pipeline from `@vercel/ncc` to `rolldown`, and migrates all packages to ES Modules (`"type": "module"`). It also upgrades TS configuration parameters to modern bundler-compatible defaults (`verbatimModuleSyntax`, `isolatedModules`, `preserve` module resolution). The bundling and executable packaging with `pkg` completes successfully, but a critical regression was introduced regarding program directory resolution for packaged environments, along with minor asset resolution issues for unbundled development workflows.

## 🔍 General Feedback

- **ESM Compilation Success:** The transition to `rolldown` and modern TypeScript configuration resolves dependencies and compiles efficiently, resulting in a cleaner output footprint.
- **Type-only Imports:** The refactor to use type-only imports correctly conforms to the newly introduced `verbatimModuleSyntax: true` flag in tsconfig.base.json.
- **Executable Packaging Compatibility:** The `pkg` target successfully packages the ESM output in Node 24 target runtimes, and the `postBuild` Windows resource/metadata edits compile correctly.

packages/server/utils/counters.ts
```diff
@@ -35,16 +35,7 @@ import {
...
- * ТАК КАК БЛЯТЬ У НАС В ЖЫЭСЕ
- * НЕ ПРИДУМАЛИ НОРМАЛЬНО ПАКЕТИРОВАТЬ ГОВНО БЕЗ ВЕБПАКА
- * ИДЕМ И ПИЛИМ КОСТЫЛИ
- * kys js!
- */
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
```
GabuTheDev, line 38: [WARNING] Applying the same asset path fallback check to `packages/server/utils/counters.ts` to support both dev (unbundled) and production (bundled) environments.

```suggestion
const pkgAssetsPath = fs.existsSync(path.join(import.meta.dirname, 'assets'))
    ? path.join(import.meta.dirname, 'assets')
    : path.join(import.meta.dirname, '../assets');
```

packages/server/utils/report.ts
```diff
@@ -112,10 +112,7 @@ export async function generateReport(instanceManager: any): Promise<Report> {
     };
 }
 
-const pkgAssetsPath =
-    'pkg' in process
-        ? path.join(__dirname, 'assets')
-        : path.join(__dirname, '../assets');
+const pkgAssetsPath = path.join(import.meta.dirname, 'assets');
```
GabuTheDev, line 136: [WARNING] Applying the same asset path fallback check to `packages/server/utils/report.ts` to support both dev (unbundled) and production (bundled) environments.

```suggestion
const pkgAssetsPath = existsSync(path.join(import.meta.dirname, 'assets'))
    ? path.join(import.meta.dirname, 'assets')
    : path.join(import.meta.dirname, '../assets');
```
*(Note: Requires importing `existsSync` from `node:fs` on line 2)*

packages/common/utils/directories.ts
storycraft: resolved in 4d18436a4371ad3c0354a21c3d7e922f0c66b29f

packages/server/router/assets.ts
storycraft: > Effectively removing all differences between dev and prod environment. It runs on bundled file + source map to match with production env. Resolving pkgAssetsPath hack and allows to use bundler post processing stuffs.

resolving as invalid

packages/server/utils/counters.ts
storycraft: > Effectively removing all differences between dev and prod environment. It runs on bundled file + source map to match with production env. Resolving pkgAssetsPath hack and allows to use bundler post processing stuffs.

resolving as invalid

packages/server/utils/report.ts
storycraft: > Effectively removing all differences between dev and prod environment. It runs on bundled file + source map to match with production env. Resolving pkgAssetsPath hack and allows to use bundler post processing stuffs.

resolving as invalid

packages/server/router/index.ts
storycraft: > Effectively removing all differences between dev and prod environment. It runs on bundled file + source map to match with production env. Resolving pkgAssetsPath hack and allows to use bundler post processing stuffs.

resolving as invalid

KotRikD (approved):

PR #657, opened 2026-07-09
----

Lorenso0:
Mistake


GabuTheDev:
Out of curiosity, what's Kumori?

Lorenso0:
> Out of curiosity, what's Kumori?

Kumori is going to be a play tracker/analyzer, it will use a modified version of tosu that has the ability to read and store the replay frame data of every single play, even failed ones.  This tool should be very useful for dedicated players to keep track of their progress and quickly analyze replays to see what went wrong.

GabuTheDev:
@Lorenso0 I reached out via osu! chat regarding some concerns. Please answer when possible.

PR #672, opened 2026-07-17
Resolve outdated static folder path creation and default to local share on Linux

GabuTheDev:
Reported on discord at https://discord.com/channels/1056534107330445362/1527660136154267800

I also changed the env default to the linux path, because imo it's kind of absurd to have `./static` in the .env but find the static folder at `~/.local/share/tosu/static`

Fallback for `./static` on linux still remains, as I am not willing to do config migration for this little inconvenience.

### DIFF
--- a/packages/common/utils/config.ts
+++ b/packages/common/utils/config.ts
@@ -2,6 +2,7 @@ import { createHash } from 'crypto';
 import * as dotenv from 'dotenv';
 import EventEmitter from 'node:events';
 import fs from 'node:fs/promises';
+import { homedir } from 'node:os';
 import path from 'path';
 
 import type {
@@ -65,7 +66,12 @@ const defaultSchema: ConfigSchema = {
     },
     staticFolderPath: {
         binding: 'STATIC_FOLDER_PATH',
-        default: './static'
+        default:
+            process.platform === 'linux'
+                ? process.env.XDG_DATA_HOME
+                    ? path.join(process.env.XDG_DATA_HOME, 'tosu', 'static')
+                    : path.join(homedir(), '.local', 'share', 'tosu', 'static')
+                : './static'
     },
     enableIngameOverlay: {
         binding: 'ENABLE_INGAME_OVERLAY',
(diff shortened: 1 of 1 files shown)


packages/common/utils/config.ts
```diff
@@ -65,7 +66,12 @@ const defaultSchema: ConfigSchema = {
     },
     staticFolderPath: {
         binding: 'STATIC_FOLDER_PATH',
-        default: './static'
+        default:
+            process.platform === 'linux'
+                ? process.env.XDG_DATA_HOME
+                    ? path.join(process.env.XDG_DATA_HOME, 'tosu', 'static')
+                    : path.join(homedir(), '.local', 'share', 'tosu', 'static')
+                : './static'
```
cyperdark, line 71: use `getStaticPath` here?

cyperdark (changes requested):

packages/common/utils/config.ts
GabuTheDev: > use `getStaticPath` here?

That makes no sense?

if `config.staticFolderPath` exists, then `getStaticPath` would return that.
Meaning that at runtime, `default == set_value`

GabuTheDev: I am guessing this is what you wanted: https://github.com/tosuapp/tosu/pull/672/commits/03ee809f658a3a7fd989216340878f2dba50888e

packages/common/utils/config.ts
```diff
@@ -492,5 +495,5 @@ export async function configInitialization() {
     await ConfigManager.initialize(configPath);
 
     // Create user-specified static folder
-    await fs.mkdir(config.staticFolderPath, { recursive: true }).catch(null);
+    await fs.mkdir(getStaticPath(), { recursive: true }).catch(null);
```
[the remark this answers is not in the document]
GabuTheDev: who tf asked your opinion???
why does github trigger automatic copilot review on my behalf

packages/common/utils/config.ts
cyperdark: yeye

cyperdark (approved):

KotRikD (approved):

PR #674, opened 2026-07-18
refactor: max achievable, fc score calculation logic

storycraft:
## Changes
* Bump lazer calc.
* Simulate maximum achievable total score for stable std pp calculation.

## Todo
* [x] Move full combo score calculation logic into separated file.
* [x] Move maximum achievable score calculation logic into separated file.
* [x] Implement maximum achievable `totalScore` calculation to account score based miss estimation in std pp calculation. Due to absence of this, maximum achievable pp was sometimes being smaller than current pp in midplay.

| before | after |
| --- | --- |
| image | image |

### DIFF
--- a/packages/tosu/src/states/gameplay.ts
+++ b/packages/tosu/src/states/gameplay.ts
@@ -14,6 +14,7 @@ import type {
 import { calculateGrade } from '@/utils/calculators';
 import { defaultCalculatedMods } from '@/utils/osuMods';
 import { type CalculateMods, OsuMods } from '@/utils/osuMods.types';
+import { calculateFcScore, calculateMaxAchievableScore } from '@/utils/score';
 
 export const defaultStatistics = {
     miss: 0,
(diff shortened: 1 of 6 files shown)


packages/tosu/src/utils/score.ts
```diff
@@ -0,0 +1,126 @@
...
+        score,
+        getMaxHit(beatmap.mode, score) + current.misses + current.comboBreaks
+    );
+    score.comboBreaks = 0;
+    score.misses = 0;
+
+    // Full combo scores doesn't have large tick and slider end misses.
+    score.largeTickHits = max.largeTickHits;
+    score.largeTickMisses = 0;
+    score.sliderEndHits = max.sliderEndHits;
```
cyperdark, line 86: i think we should make FC to use replay values instead of using max values. Because if player fc's a map with their tick and slider end misses they will see different pp on tosu than on result screen

storycraft: They affect combo. And missing them will increase internal miss count(aka effective miss count).

Proper full combo pp should be higher than scores with slider end, tick misses imo. Btw this is cleaned up code of original implementation in tosu so nothing is changed.

packages/tosu/src/states/gameplay.ts
```diff
@@ -554,71 +555,29 @@ export class Gameplay extends AbstractState {
...
-                calcOptions.greats = this.statistics.great;
-            }
-            calcOptions.accuracy =
-                currentBeatmap.calculateAccuracy(calcOptions);
-
             const maxAchievablePerformance =
                 currentBeatmap.calculatePerformance(
                     beatmapPP.difficultyAttributes,
-                    calcOptions
+                    calculateMaxAchievableScore(
```
GabuTheDev, line 561: Why do these new functions end with `Score` when this is about `pp` ??

storycraft: They are about score, not pp. `calculatePerformance` *is* about pp.

packages/tosu/src/utils/score.ts
cyperdark: okay

KotRikD (approved):

PR #677, opened 2026-07-25
fix: add scorev2 mod on multi

### LINKED ISSUE #418: The `ScoreV2` mod is not being detected in Multiplayer.
Scorev2 enabled in MP relies on Win Condition, which is not included in the mods in the API. Therefore, some functions I want to implement/display only when V2 is enabled will not work in MP. Can we add corresponding functions so that V2 can be detected in MP?

storycraft:
* Manually add scorev2 mod when scorev2 score processor exists.
* Fix #418


GabuTheDev (approved):
Code looks 👍

GabuTheDev:
I would've wished to have win condition and room type read, and provided via ws.
And as a side effect of that, to have this fixed 
I guess https://github.com/tosuapp/tosu/pull/650 could fix that, but it only touches lazer for now.

KotRikD (approved):

PR #678, opened 2026-07-26
feat: add hit windows to tosu websocket.

### LINKED ISSUE #621: Provide `hit_error_bar`  boundries & hit error anchor
this arrow and bar 

Image

adding this just to not forget about it in future

storycraft:
* Introduce `hitWindow` object in v2 json
* ~Blocked by #674~
* Close #621

## Examples

std (with HDDTCL)
```json
{
...
  "beatmap": {
    "hitWindow": {
      "miss": 266.666666666667,
      "meh": 83,
      "ok": 53,
      "great": 23
    },
    ...
  },
...
}
```

mania (CL)
```json
{
...
  "beatmap": {
    "hitWindow": {
      "miss": 162.5,
      "meh": 125.5,
      "ok": 101.5,
      "good": 71.5,
      "great": 38.5,
      "perfect": 16.5
    },
    ...
  },
...
}
```

taiko (CL)
```json
{
...
  "beatmap": {
    "hitWindow": {
      "miss": 84.5,
      "ok": 67.5,
      "great": 28.5
    },
    ...
  },
...
}
```


storycraft:
Only last commit is part the pr and rests are #674

GabuTheDev:
Is this reviewable? It's still in draft.

storycraft:
> Is this reviewable? It's still in draft.

yeah

packages/tosu/src/api/types/v2.ts
```diff
@@ -238,6 +240,10 @@ export interface Beatmap {
     stats: Stats;
 }
 
+export type HitWindow = Partial<
+    Record<Uncapitalize<keyof typeof OsuHitResult>, number>
+>;
```
cyperdark, line 245: i would prefer if all values will be there by default, instead of having Partial and undefined in overlays json

Image


```ts
const thing = (data.beatmap.stats.hitWindow.perfect || 0) + (data.beatmap.stats.hitWindow.perfect || 0)
```

packages/tosu/src/states/beatmap.ts
```diff
@@ -570,9 +571,15 @@ export class BeatmapPP extends AbstractState {
...
-                hitWindow: HitWindows.getGreatHitWindow(
-                    this.beatmap.mode,
-                    convertedDifficulty.overallDifficulty
+                hitWindow: Object.fromEntries(
+                    Array.from(hitWindows.allAvailableWindows()).map(
+                        ([key, value]) => [
+                            OsuHitResult[key].replace(/^./, (s) =>
+                                s.toLowerCase()
+                            ),
+                            value / this.clockRate
```
cyperdark: what about moving `Array.from(hitWindows.allAvailableWindows())` to hitWindows const?

```ts
const hitWindows = Array.from(this.beatmap.createHitWindows().allAvailableWindows());
```

cyperdark (changes requested):

packages/tosu/src/api/types/v2.ts
storycraft: The purpose of the Hit window is to provide available hit result types and their timing information. If I add a fallback value to all types, wouldn't users have to write their own logic to select hit results based on the game mode?

I forgot to include but this is example json of catch mode
```json
{
...
  "beatmap": {
    "hitWindow": {
      "miss": 0,
      "meh": 0,
      "ok": 0,
      "good": 0,
      "great": 0,
      "perfect": 0
    },
    ...
  },
...
```
with no timing ms.

packages/tosu/src/states/beatmap.ts
storycraft: yeah

GabuTheDev (approved):

packages/tosu/src/api/types/v2.ts
cyperdark: @KotRikD what do you think?

KotRikD: At one side I'm okay with partial values, at the second users will forget checking other gamemods, and will use just std one :(

cyperdark (approved):

PR #685, opened 2026-08-05
Bump dependencies

KotRikD:
PR made using AI tools

Claude Opus 5


GabuTheDev:
Did you really use Opus 5 to bump some dependencies????????????????????????????????????????? bruh

cyperdark:
needed review from @storycraft, since there was a bump of electron dependencies which used by ingame overlay

packages/ingame-overlay/package.json
```diff
@@ -18,11 +18,12 @@
   },
   "devDependencies": {
     "@types/node": "^24.10.4",
-    "electron": "^39.2.7",
-    "electron-builder": "^26.4.0",
+    "electron": "^40.1.0",
```
storycraft: Electron overlay renderer expect specific version of Electron(defined as peerDependencies). They can introduce breaking changes between major versions. In fact, some previous versions did before.

storycraft (changes requested):

storycraft:
> needed review from @storycraft, since there was a bump of electron dependencies which used by ingame overlay

Yes thank you for noticing this. Because it actually does matter.

packages/ingame-overlay/package.json
storycraft: I didn't test the build. But if it works without any issues, you can resolve this and ignore.

KotRikD:
> Did you really use Opus 5 to bump some dependencies????????????????????????????????????????? bruh

`¯\_(ツ)_/¯` claude sub, why not

image

PR #689, opened 2026-08-14
fix: restore broken star history chart

OctoBored:
The star history chart in the README is currently broken because the GitHub stargazer API used to render it is rate-limited. This change switches the chart to a working alternative so it displays correctly again.


GabuTheDev:
We'd rather wait for GitHub to figure this out (and still use the well-known star-history service), than use some unknown "alternative" service.

Thank you for your contribution though!

PR #693, opened 2026-08-23
fix: normalize SC tokens for tosu websocket

KotRikD:
Fixes SC token behavior

Grok 4.6 [xhigh]


KotRikD:
It's not breaking, its not worked as expected at all

PR #708, opened 2026-09-06
fix(stable): allow falsy config and keybind values in settings

### LINKED ISSUE #707: stable: falsy config values are never applied
1. Describe your issue as detailed as possible

on stable, config values that read as falsy never make it into the api. once `settings.mouse.rawInput` (or `resolution.fullscreen`, `mouse.disableButtons`, any of them) has been true once while tosu is running, it stays true until i restart tosu, no matter what i set in osu

repro of one of the possible cases:
- raw input ON in osu, start tosu -> `/json/v2` says `settings.mouse.rawInput: true` (correct)
- turn raw input OFF -> still `true` (wrong, stays like that forever)

(starting with it OFF looks correct but isn't actually reading anything since `Settings.mouse.rawInput` just defaults to `false` in `states/settings.ts`)

the cause is in [packages/tosu/src/memory/stable.ts#L1384](https://github.com/tosuapp/tosu/blob/4ec3755a0cb13259778050cf8f964eca11acfab5/packages/tosu/src/memory/stable.ts#L1384) in `settings()` on current master:

```ts
if (result === null || !result.key || !result.value)
    continue;
```

`configValue()` decodes bools with `Boolean(readByte(...))` so a false comes back as an actual `false`, and `!result.value` skips it. `configValue()` already returns `null` on a bad read, so the caller re-checking the value itself is just wrong.

it's not only bools, everything that can be falsy gets dropped so like:
- `bool` -> `false`
- `int` / `double` (`readDouble`) -> `0`, so volume sliders, `background.blur`, `background.dim`, `audio.offset.universal`, `mouse.sensitivity` etc can never actually reach 0 in the api, they report the last non-zero value
- `enum` (`readInt`) -> `0`
- `string` / `bstring` -> `""`

and it sticks because `settings()` returns a fresh object but `Settings.updateState()` only `setNestedValue`s the keys that are present, onto a long-lived state object. skipped key means that the last value stays

same guard is on [the bindings loop at line 1417](https://github.com/tosuapp/tosu/blob/4ec3755a0cb13259778050cf8f964eca11acfab5/packages/tosu/src/memory/stable.ts#L1417) so an unbound key never updates either. safe to drop there too, `VirtualKeyCode.None = 0` exists so `VirtualKeyCode[0]` is `'None'`, not undefined.

should be `result.value === undefined` (or `== null`) instead of `!result.value`. `isAllowedValue()` below is type-only (`typeof value === 'boolean'` / `isRealNumber`) so `false` and `0` pass it fine, no other change needed im pretty sure

2. tosu.exe location

H:\Softwares\tosu

3. Record video

no video, the repro above is 2 toggles and a curl to `/json/v2` so prolly easier to follow than a recording. can record one if you actually want it tho

4. Windows or Linux

windows 11
tosu version: v4.26.2

GabuTheDev:
Resolves https://github.com/tosuapp/tosu/issues/707

### Summary of Changes
- Replaced truthiness checks (`!result.value`) in `StableMemory.settings()` with explicit `undefined` and `null` checks (`result.value === undefined || result.value === null`) for both configuration and keybinding value reads.

### Problem Solved
On osu! stable, config values that evaluated to falsy (`false` for booleans, `0` for numbers, `""` for strings, and `0` / `VirtualKeyCode.None` for keybinds) were skipped in `StableMemory.settings()`. Because `Settings.updateState()` only updates keys that are present in the returned dictionary, once a setting held a non-falsy value, it could never update back to `false`, `0`, or `None`.

### Files Affected
- `packages/tosu/src/memory/stable.ts`

### Testing & Verification
- Created an isolated unit test reproducing the issue using mock memory reads for `RawInput: false`, `VolumeUniversal: 0`, `Skin: ""` and unbound keybinding `TaikoInnerLeft: 0` (`VirtualKeyCode.None`).
- Verified that the original logic failed the assertions and the updated logic passed all checks.
- Verified TypeScript compilation via `pnpm run -C packages/tosu ts:compile`.
- Verified formatting and linting via `pnpm run prettier:ci` and `pnpm run lint:ci`.

***

Fixed by Gemini 3.8 Flash

### DIFF
--- a/packages/tosu/src/memory/stable.ts
+++ b/packages/tosu/src/memory/stable.ts
@@ -1381,7 +1381,12 @@ export class StableMemory extends AbstractMemory<OsuPatternData> {
                 try {
                     const result = this.configValue(configPointer, position);
                     if (result instanceof Error) throw result;
-                    if (result === null || !result.key || !result.value)
+                    if (
+                        result === null ||
+                        !result.key ||
+                        result.value === undefined ||
+                        result.value === null
+                    )
                         continue;
 
                     const value = configList[result.key];
@@ -1414,7 +1419,12 @@ export class StableMemory extends AbstractMemory<OsuPatternData> {
                 try {
                     const result = this.bindingValue(bindingPointer, position);
                     if (result instanceof Error) throw result;
-                    if (!result.key || !result.value) continue;
+                    if (
+                        !result.key ||
+                        result.value === undefined ||
+                        result.value === null
+                    )
+                        continue;
 
                     const value = bindingList[result.key];
                     if (!value || !isAllowedValue(value[0], result.value))


cyperdark (approved):

KotRikD:
@codex review

